PR #3, opened 2023-06-01
Initial import of the ingest command.

calebbrown:
Largely feature complete implementation of OSV ingestion command:

- YAML config file
- multiple sources
- saves progress for speedy imports
- "lookback" to allow for missed writes that may have been in-flight
- atomic file writing to avoid corruption
- saves "origin" information in a database specific section
- sanitizes and validates paths generated from package

Tests are there, but coverage is still being improved.


cmd/ingest/source/source.go
```diff
@@ -0,0 +1,59 @@
...
+var (
+	ErrInvalidID  = errors.New("invalid source id")
+	validIDRegExp = regexp.MustCompile("[a-z0-9-]+")
+)
+
+type Source struct {
+	ID              string `yaml:"id"`
+	Bucket          string `yaml:"bucket"`
+	Prefix          string `yaml:"prefix"`
+	LookbackEntries int    `yaml:"lookback-entries"`
```
oliverchang, line 34: Could you explain why this needs to be configurable rather than a constant?

oliverchang (approved):
nice!

cmd/ingest/source/source.go
calebbrown: That's a fair point. It could be a constant for sure.

I was hoping to eventually have an end-to-end test for the tool to ensure it is working correctly so an external config would be useful.

But yes, it could be hard-coded.

calebbrown: I might leave it as is. I do find it useful for local development.

oliverchang: ack, fair enough.

cmd/ingest/main.go
```diff
@@ -0,0 +1,210 @@
...
+		keys.Set(s.ID, end)
+	}
+
+	// Atomically write updated start keys...
+	if err := saveStartKeys(*startKeysFlag, keys); err != nil {
+		log.Fatalf("Failed saving start keys: %v", err)
+	}
+
+	// Clean up temp directory
+	if err := os.RemoveAll(tempDir); err != nil {
```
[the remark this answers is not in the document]
calebbrown: Done

oliverchang (approved):

PR #29, opened 2023-06-19
Introduce the preprocess command that prepares the repo for ID allocation

calebbrown:
Preprocess updates OSV reports ready for ID allocation.

- Normalizes the Summary
- removes DatabaseSpecific values that aren't allowed
- removes EcosystemSpecific and Severity values
- merges reports!

This PR also adds lots of tests to ensure the merge behavior works as expected, but more tests are still to come.


oliverchang (approved):
very nice!

PR #34, opened 2023-06-22
Add more license and an initial CONTRIBUTING doc

calebbrown:


cmd/preprocess/main.go
```diff
@@ -1,3 +1,17 @@
+// Copyright 2022 Malicious Packages Authors
```
oliverchang, line 1: nit: 2023? :)

oliverchang (approved):

cmd/preprocess/main.go
calebbrown: lol. fixed. everywhere.

PR #36, opened 2023-06-23
Allow local OSV ingestion from the command line.

calebbrown:
Adds -source and -dir flags that allow someone to ingest OSV from a local directory.

`-source` can be a source in the config, or, if `-dir` is specified, it can be arbitrary.

If `-dir` is set `-source` must be set as well.

Local directory ingestion uses gocloud's fileblob to support ingestion.

### DIFF
--- a/cmd/ingest/main.go
+++ b/cmd/ingest/main.go
@@ -77,35 +77,26 @@ func main() {
 	}
 
 	// Determine sources based on flags
-	var sources []*source.Source
+	sources := c.Sources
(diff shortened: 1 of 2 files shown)


oliverchang (approved):
nice! is there any documentation on how to use this?

calebbrown:
Docs are coming soon. I'll push something up next week that covers the different types of contributions.

I'm currently implementing a basic validator to ensure contributions go in the right location.

PR #37, opened 2023-06-23
Add a command for doing basic validation on the repo.

calebbrown:
Does the following checks:

- path based on ecosystem and package name matches the actual path, and is valid
- report is parsable as JSON
- ID conforms to the correct format
- ID (if assigned) matches the filename, excluding extension

### DIFF
--- a/cmd/validate/main.go
+++ b/cmd/validate/main.go
@@ -1,0 +1,139 @@
+// Copyright 2023 Malicious Packages Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


cmd/validate/main.go
```diff
@@ -0,0 +1,139 @@
...
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/ossf/malicious-packages/internal/config"
+	"github.com/ossf/malicious-packages/internal/report"
+	"github.com/ossf/malicious-packages/internal/reportio"
+)
+
+var validIDRe = regexp.MustCompile(`^([A-Z]+)-[0-9]{4}-[0-9]+$`)
```
oliverchang, line 32: Should we just fix this to an explicit "MAL-" prefix?

calebbrown: I've tried to keep the MAL prefix specified via the config across the commands, with a view to making it easy to support anyone who wants to fork the repo.

It is checked in the function where the regexp is used.

oliverchang (approved):

PR #49, opened 2023-06-29
Add contributing details

calebbrown:


CONTRIBUTING.md
```diff
@@ -9,11 +9,146 @@ We'd love to accept your patches and contributions to this project.
...
+    - From the repo run (requires [Go](https://go.dev/doc/install)):
+
+    ```shell
+    go run ./cmd/ingest \
+       -config config/config.yaml \
+       -dir /path/to/osv \
+       -source [source id]
+    ```
+
+    - Fix any issues. You will need to revert the repo before running
```
oliverchang, line 87: Nit: perhaps we can avoid overloading the use of "issues" here and use an alternative like "problems reported" or something.

CONTRIBUTING.md
```diff
@@ -9,11 +9,146 @@ We'd love to accept your patches and contributions to this project.
...
+      ingest again.
+1. Commit and push the changes to a form and submit a Pull Request.
+    - Reference the issue from step #1 in the PR.
```
oliverchang, line 90: What if someone wants to do periodic bulk imports? Would they reference the same issue each time?  And that issue would stay open?

oliverchang (commented):
nice!

CONTRIBUTING.md
calebbrown: Fixed.

CONTRIBUTING.md
calebbrown: I've tried to clarify this. Let me know what you think.

oliverchang (approved):

calebbrown:
Ping @jossef

Issue #61, opened 2023-07-11
Ingest GHSAs for malicious packages


oliverchang:
The official GHSA OSV database (https://github.com/github/advisory-database) currently does not include malware entries suchc as https://github.com/advisories/GHSA-xv2f-5jw4-v95m, and it's unclear if they will be doing this in the short/medium term.

We could leverage existing GHSA->OSV code in https://github.com/ossf/osv-schema/tree/main/tools/ghsa to scrape the GraphQL and import them into this repo to fill that gap.

Some open questions:
- ~~Do we still still assign a MAL- ID for these? If we do, and GitHub starts exporting them later, it may cause a bit of confusion. This could be potentially handled via OSV `alias`es though.~~ thinking more it  fits our merging/de-duplication workflow a lot better to assign a MAL- entry.

calebbrown:
The backfill has been done.

Once I confirm the automated ingestion is working, I'll update the GH workflow to enable committing.

PR #68, opened 2023-07-19
Add an option to copy an ID into aliases. Also support no-bucket sources

calebbrown:


calebbrown:
This is related to #61

oliverchang (approved):

PR #89, opened 2023-07-31
Use `pipenv run` instead of `pipenv shell`

calebbrown:
`pipenv shell` doesn't work because it expects an interactive TTY.


oliverchang (approved):

calebbrown:
See #61

PR #105, opened 2023-08-10
Ingest package analysis OSV data

calebbrown:
Adds a GitHub workflow for automatically ingesting content from cloud storage sources.

Adds a config entry for ingesting Package Analysis detection OSV content


adaluong (approved):
lgtm!

PR #107, opened 2023-08-10
Add a hack to bypass a bug in GoCloud's anonymous GCS support.

calebbrown:
GoCloud's GCS blob implementation is meant to support anonymous access using an `access_id=-` parameter.

However, in the Lazy URL opener (https://github.com/google/go-cloud/blob/v0.33.0/blob/gcsblob/gcsblob.go#L153), `gcp.DefaultCredentials(ctx)` is always called, which results in an error when a credential file is not present.

This bug works around this by providing the `golang.org/x/oauth/google` library internally with the minimum JSON object required to successfully return a valid credential object.

### DIFF
--- a/.github/workflows/ingest-cloud.yml
+++ b/.github/workflows/ingest-cloud.yml
@@ -25,9 +25,17 @@ jobs:
       with:
         go-version: "1.20"
 
+    # gocloud has an issue where it requires an ADC file to work with GCS, even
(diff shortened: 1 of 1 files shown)


calebbrown:
This is a work around for https://github.com/google/go-cloud/issues/3302

oliverchang (approved):

PR #135, opened 2023-08-25
Fix capitalization of PyPI in ecosystem field.

calebbrown:
Turns out ecosystem is case sensitive and it should be "PyPI" instead of "pypi".


calebbrown:
For future reference this is the command I used:

```bash
$ find . -name "MAL-*.json" | xargs -Iosv grep -l "\"ecosystem\": \"pypi\"" osv | xargs -Iosv sed -i -e 's|\"ecosystem\": \"pypi\"|\"ecosystem\": \"PyPI\"|' osv
```

PR #177, opened 2023-09-28
Refactor vuln validation and validate ranges.

calebbrown:
Also adds testing to ensure it functions correctly.

This change will ensure OSV conforms to the spec, and allows simpler processing logic by allowing assumptions to be made about the correctness of the OSV.


internal/report/validate.go
```diff
@@ -0,0 +1,101 @@
...
+		if c == 0 {
+			return fmt.Errorf("%w: no event type is specified", ErrInvalidOSV)
+		}
+		if c > 1 {
+			return fmt.Errorf("%w: more than one event type is specified", ErrInvalidOSV)
+		}
+	}
+	// Entries in the events array can contain either last_affected or fixed
+	// events, but not both.
+	if hasFixed && hasLastAffected {
```
oliverchang, line 93: FYI: there's some overlap with the JSON schema validation here: https://github.com/ossf/osv-schema/blob/main/validation/schema.json.

Might make sense in the future to re-use some of that instead.

oliverchang (approved):

internal/report/validate.go
calebbrown: True. Definitely a good idea moving forward.

PR #193, opened 2023-10-12
Add an initail schema doc, explaining the current database_specific use

calebbrown:


docs/schema_additions.md
```diff
@@ -0,0 +1,132 @@
...
+		/* ... other database specific entries ... */
+    }
+}
+```
+
+# Field Details
+
+## malicious-packages-origins fields
+
+The `malicious-packages-origins` field is a JSON array containing objects that
```
oliverchang, line 54: Nit: should we state up front that this is likely not useful for most consumers, and is more internal state for the malicious-packages repo? 

Also consider moving this to the last section as well, as otherwise readers might get lost in this before they get to the ones that are more interesting to them.

docs/schema_additions.md
```diff
@@ -29,6 +30,11 @@ elaborated in the next section. All strings contain UTF-8 text.
 {
     /* ... other OSV fields ... */
     "database_specific": {
+		"iocs": {
```
oliverchang, line 33: nit: is the indentation inconsistent here with line 38?

docs/schema_additions.md
calebbrown: I moved it lower and updated the first paragraph to indicate more clearly it is for internal use.

docs/schema_additions.md
calebbrown: Fixed.

oliverchang (approved):

PR #206, opened 2023-10-19
Ensure there are no duplicate ranges when merging.

calebbrown:
Currently ranges are just appended naively together.

This change will ensure that if two or more ranges that are identical are added, then only one of them is kept.

### DIFF
--- a/internal/report/merge.go
+++ b/internal/report/merge.go
@@ -17,6 +17,7 @@ package report
 import (
 	"errors"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/google/osv-scanner/pkg/models"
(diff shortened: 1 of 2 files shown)


internal/report/merge.go
```diff
@@ -140,6 +141,59 @@ func combineCredits(creditSets ...[]models.Credit) []models.Credit {
...
+	}
+	// Either we don't have any database specific entries, or they are entirely equal.
+	return (len(r1.DatabaseSpecific) == 0 && len(r2.DatabaseSpecific) == 0) || reflect.DeepEqual(r1, r2)
+}
+
+// combineRanges combines and reduces one or more range slices.
+//
+// The function assumes the input ranges are valid OSV ranges. If two or more
+// ranges are identical, only one will be included.
+func combineRanges(rangeSets ...[]models.Range) []models.Range {
```
oliverchang, line 183: Would it make sense to add some logic to detect cases like 

```
{
  "introduced": "0"
}
```

And have this consume and remove any other ranges, given that `introduced: 0` on its own expresses every single version in existence is affected.

oliverchang (approved):

internal/report/merge.go
calebbrown: I did consider doing that, but there are some considerations that led to me deferring the work.

For example, how do we handle merging two OSV where one is overly broad and uses `{"introduced": "0"}`, but the other is more specific and accurate?

PR #257, opened 2023-11-27
[fix] OSV does not support range SEMVER for ecosystem PyPI.

calebbrown:
This change updates the ranges so they are correct, and adds validation to prevent this from occurring again.


calebbrown:
@jossef just an FYI - the range "type" needs to be "ECOSYSTEM" instead of "SEMVER" for PyPI.

oliverchang (approved):

PR #381, opened 2024-03-04
Clarify what is allowed and what is considered malicious.

calebbrown:
Further clarify the specification of what is considered malicious for the repository.

This helps make decisions about true vs false positives easier to make.


README.md
```diff
@@ -29,30 +29,104 @@ These public reports help protect the open source community, and provide a data
...
-- https://github.com/lxyeternal/pypi_malregistry (PyPI)
-- https://dasfreak.github.io/Backstabbers-Knife-Collection/ (PyPI and npm), by Marc Ohm et al.
-- https://github.com/datadog/malicious-software-packages-dataset (PyPI), by Datadog
+Below is the definition of what this repository considers a malicious package.
+
+- an open source package publicly available in a package registry
+- and either:
+  - when installed or used, would require some sort of incident response; or
+  - exfiltrates an identifier that can be directly used to launch an attack
+    against the victim (e.g. username for phishing or password bruteforcing)
```
steiza, line 64: Minor: you don't have to list out all the ways malicious software behaves, but these days I would say stealing browser cookies is pretty high on the list, maybe also installing a keylogger

README.md
```diff
@@ -29,30 +29,104 @@ These public reports help protect the open source community, and provide a data
...
 
-- GitHub's [Advisory Database (filtered by malware)](https://github.com/advisories?query=type%3Amalware), for the NPM ecosystem.
-- https://github.com/lxyeternal/pypi_malregistry (PyPI)
-- https://dasfreak.github.io/Backstabbers-Knife-Collection/ (PyPI and npm), by Marc Ohm et al.
-- https://github.com/datadog/malicious-software-packages-dataset (PyPI), by Datadog
+Below is the definition of what this repository considers a malicious package.
+
+- an open source package publicly available in a package registry
+- and either:
+  - when installed or used, would require some sort of incident response; or
```
steiza, line 62: Minor: definitions are hard, and I wouldn't call myself an expert here, but maybe something like "when installed or runs, tries to persist software on the machine unrelated to the advertised function of the package"

steiza (approved):
Nice - protestware is a good call-out as well

SecurityCRob (approved):
This is a great update.  As I have cycles, I'll see if I have any wording bikeshedding I could contribute, but I think this is mvp as is

README.md
```diff
@@ -29,30 +29,116 @@ These public reports help protect the open source community, and provide a data
...
+repository.
+
+### Reverse engineering protection (e.g. obfuscation)
+
+Reverse engineering protections are not malicious, unless it exhibits malicious
+behavior as-per the definition above.
+
+Obfuscation, debugger evasion, and other reverse engineering protection
+techniques, are used by both developers seeking to protect their source code
+and attackers seeking to evade detection.
```
bjorn3, line 108: The crates.io usage policy forbids any content that

> uses obfuscation to hide or mask functionality

which seems to suggest to me that even using obfuscation for protecting source code is considered unacceptable by crates.io, not just if the obfuscation is done to hide malicious behavior.

calebbrown:
Merged PR since there has been no activity and after CRob's review.

Thanks all!

PR #421, opened 2024-03-31
Add bugsnagmw

lukehinds:
Add report for bugsnagmw. This package has been taken down from npm registry after we reported our findings.


osv/malicious/npm/bugsnagmw/MAL-0000-bugsnagmw.json
```diff
@@ -0,0 +1,45 @@
...
+        "source": "trustypkg.dev",
+        "versions": [
+          "1.0.0",
+          "1.0.1",
+          "1.0.2",
+          "1.0.3"
+        ]
+      }
+    ]
+  }
```
calebbrown, line 44: This section isn't required if this is a hand-crafted OSV file.

It's largely there to manage ingestion from a bulk collection of upstream OSV files to detect whether or not the OSV has been ingested before.

Without the sha256 this will do weird things to the details section.

Finally if you do wish to do bulk or automated ingestion I am happy to agree on a source ID that works for you.

calebbrown (commented):
Thanks for the submission!

I just have a minor comment about the origins part.

calebbrown (approved):

PR #428, opened 2024-04-05
Manually skip withdrawn for @m365-admin/customizations as the package was malicious.

calebbrown:
- ref: https://github.com/advisories/GHSA-8646-rr57-mhf6
- ref: https://socket.dev/npm/package/@m365-admin/customizations


calebbrown:
Added versions to ensure that if the range needs to be removed there is some version detail present.

maxfisher-g (approved):

PR #478, opened 2024-05-09
Add support for multiple prefixes per source.

calebbrown:
This allows for more diverse types of sources, such as when multiple streams of OSV files may be present in a single bucket under different prefixes.


oliverchang (approved):

oliverchang:
(whoops, accidentally merged, sorry if that wasn't OK!)

PR #518, opened 2024-06-05
Add reversing-labs as a source, and allow access to AWS.

calebbrown:
Enable Reversing Labs as a source for Malicious Packages


config/config.yaml
```diff
@@ -9,3 +9,9 @@ sources:
   - "confident/"
 - id: "ghsa-malware"
   alias-id: true
+- id: "reversing-labs"
+  bucket: "s3://rl-github-ossf?awssdk=v2&profile=revlab&region=us-east-1"
+  prefixes:
+  - "RLMA-"
+  - "RLUP-"
```
rhalar, line 16: We actually use `RLUA-` for the update prefix.
Sorry for the confusion, I might have provided an example which had the RLUP prefix way back but we landed on `RLUA` in the end, which should have been communicated at some point I believe.

calebbrown: Thanks, I've updated this.

rhalar (commented):
Prefixes seem fine now, should be good to go!

.github/workflows/ingest-cloud.yml
```diff
@@ -31,6 +31,20 @@ jobs:
...
+    # Setup v2 credentials for accessing AWS S3 buckets. Each set of credentials
+    # is given its own profile which is then set in the config using the
+    # "profile=PROFILENAME" query parameter in the source's bucket URL.
+    - name: AWS Credentials
+      run: |
+        mkdir "$HOME/.aws"
+        echo '[reversinglabs]' > "$HOME/.aws/credentials"
+        echo "aws_access_key_id = $RL_AWS_ACCESS_KEY_ID" >> "$HOME/.aws/credentials"
+        echo "aws_secret_access_key = $RL_AWS_SECRET_ACCESS_KEY" >> "$HOME/.aws/credentials"
+        echo '[profile reversinglabs]' > $HOME/.aws/config"
```
spencerschrock, line 44: Is there any concern about credential leaking? I know GitHub will hide this in the logs, but this goes against their recommendations:
https://docs.github.com/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow


> Avoid passing secrets between processes from the command line, whenever possible. Command-line processes may be visible to other users (using the ps command) or captured by [security audit events](https://docs.microsoft.com/windows-server/identity/ad-ds/manage/component-updates/command-line-process-auditing). To help protect secrets, consider using environment variables, STDIN, or other mechanisms supported by the target process.

calebbrown: Fixed. Now uses `envsubst` to generate the file.

maxfisher-g (approved):

PR #554, opened 2024-06-28
Reverse incorrectly created malicious entry for package React.

lujunsan:
The malicious package is @zettle-bo/react instead, which was already published as of 2023-09-21T01:41:32Z.

Not sure what the correct procedure is here, please advise.

This has happened to multiple packages, the NPM scope is not being taken into account as it is flagging the real, non-malicious packages as malicious.


calebbrown:
Created #555 to track this issue.

calebbrown:
For this report specifically, it should be marked as withdrawn.

This involves moving it to the correct osv/withdrawn/ location and adding the "withdrawn" time to the OSV.

calebbrown:
This is now fixed in PR #559.

See https://github.com/ossf/malicious-packages/blob/main/osv/withdrawn/npm/react/MAL-2024-2929.json

Issue #555, opened 2024-06-28
NPM packages with namespaces from ReversingLabs don't have namespace in name.


calebbrown:
As at the time of reporting the number is 232 in NPM.

For example `@zettle-bo/react` is listed as `react`.

calebbrown:
I will pause the reversing labs ingestion to ensure no new data is ingested until we have confirmed a fix.

Errant reports can either be corrected (moved to the correct location) in the case that the report doesn't yet exist, or withdrawn if the do.

calebbrown:
@rhalar

G-Rath:
fwiw it looks like https://osv.dev/vulnerability/MAL-2024-3239 is also a result of this - it would be good to know how we can get these invalid advisories cleared since they're being flagged by the scanner.

I also wonder if osv.dev should be doing some validation around the PURLs as it sounds like they're invalid so maybe the `affected` field or the whole advisory be rejected rather than ingested?

calebbrown:
False positives and data problems have been resolved. Some Reversing Labs reports need to be re-added, but that is less urgent.

rhalar:
Hey,

to clarify for anyone affected or following this issue, I work for ReversingLabs and am 'in charge' for exposing our advisories, and as such responsible for this mess :)
I contacted Caleb and we will resolve the source of the problem very soon, to avoid any further problems with the data we expose.

I'd like to apologize for any alarms this may have triggered, it was a dumb oversight we should have caught, and I take the full blame for it.
A huge shutout and thanks to @calebbrown for doing damage control and reacting so quickly.

Our security researchers work hard and do a great job at finding malicious packages across multiple ecosystems, I hate to see their work besmirched by what is essentially a conversion error by my side; we'll make sure this doesn't happen again.

It's not the best first impression but we are very excited by this collaboration, and hope to continually contribute to the OSSF going forward.

calebbrown:
The data fixing is now complete. I will close this issue once I have un-paused the reversing-labs source in the config.

lujunsan:
Hi @calebbrown . It looks like some packages are missing from the "withdrawn" list, such as NPM `qs` ([osv](https://osv.dev/vulnerability/MAL-2024-2909)) or NPM `angular` ([osv](https://osv.dev/vulnerability/MAL-2024-1755)). Both of these were reverted because of this issue but they were not added to the withdrawn directory, which is causing some inconsistencies in our systems.

calebbrown:
[...]

PR #557, opened 2024-07-01
Ensure purls are consistent and valid if they are present.

calebbrown:
This helps prevent and detect cases like #555.

This PR will have failing checks until the problems in #555 are dealt with.

osv-scanner is updated in this PR too.


go.mod
```diff
@@ -1,12 +1,14 @@
 module github.com/ossf/malicious-packages
 
-go 1.21
+go 1.21.11
+
+toolchain go1.22.4
```
maxfisher-g, line 5: Are the toolchain directive and go patch version number changes intentional?

maxfisher-g (approved):

go.mod
calebbrown: Not really. They're from updating osv-scanner.

For what it's worth they don't make a whole lot of difference to this project, so I am fine with them.

PR #558, opened 2024-07-01
Repair NPM namespace issues after Reversing Labs import

### LINKED ISSUE #555: NPM packages with namespaces from ReversingLabs don't have namespace in name.

calebbrown:
Fixes #555.

For reports for previously unreported NPM packages, the reports have been moved to the correct location.

TBD: reports that overlap with existing NPM packages


calebbrown:
This PR is broken. I discovered some issues with my existing approach, and I need to repeat this work.

PR #602, opened 2024-07-30
Add version 4.1.2 of malicious package harthat-chain

calebbrown:
According to data we have collected version 4.1.2 was also published about a day earlier than 4.2.3.


calebbrown:
cc @poppysec

maxfisher-g (approved):

PR #667, opened 2024-10-16
Re-enable Reversing Labs.

### LINKED ISSUE #555: NPM packages with namespaces from ReversingLabs don't have namespace in name.

calebbrown:
- Will reprocesses all data.
- Based on analysis, there was one minor issue in the data. The bad data has been removed and a reprocess will add the missing report for alloy-icons (npm).

Fixes #555


calebbrown:
@rhalar

rhalar (approved):
Thanks Caleb! We are ready to continue, there is a set of new advisories on the bucket.

Beware however, if you are reprocessing the whole bucket, we did fix the NPM namespace issues in our advisories as well, so I'm guessing they will gave a different hash, if that presents a problem?
We will also do any necessary withdrawals after this is re-enabled. I think perhaps one minor update will have to be done.

PFCM:
Is there a way to see a dry run of what the ultimate changes to the repo would look like?

calebbrown:
I have run a dry run locally, and everything looks okay. I see ~3820 updates, 760 new reports.

@rhalar I just wanted to confirm that adding the `"related": [ "RLMA-2024-xxxxx" ]` entries is intentional, and that exposing these in the official OSV part of the report is intentional - or if it might be better to remove these?

PFCM (approved):

mgdcvetko:
Hi, I work with @rhalar on this project so I will try to answer your questions as he is offline for a few days. 

We do intentionally use the `related` field to connect update advisory to the original advisory for a malicious package, but we were under the impression that it won't be exposed and that it's on us to use it if we want and how we want. It does not provide any info in the official report so we would rather not have it in. Is there a way in which you can omit the `related` field?

calebbrown:
Thanks for the update @mgdcvetko. Please take a look at #671. That PR will filter out the related field.

PR #668, opened 2024-10-16
Add reports for ethers packages

poppysec:


calebbrown (approved):

calebbrown:
Great find. That is a sneaky bit of malware.

PR #671, opened 2024-10-22
Add the ability to remove an entry from the aliases or related fields.

calebbrown:
Introduces the concept of "filters", currently only including one that removes an entry from the "aliases" or "related" field based on a pattern.

This is a minimal implementation that can be expanded in the future to include other types of filters (e.g. replace).

For now this change is to support #667


internal/reportfilter/filter.go
```diff
@@ -0,0 +1,69 @@
...
+func (rf *removeFilter) Apply(v *models.Vulnerability) {
+	switch rf.field {
+	case "aliases":
+		v.Aliases = slices.DeleteFunc(v.Aliases, rf.re.MatchString)
+	case "related":
+		v.Related = slices.DeleteFunc(v.Related, rf.re.MatchString)
+	}
+}
+
+// Filters is a slice of filters that itself implements the Filter interface.
```
PFCM, line 62: maybe mention this applies them in order, just to be pedantic

PR #673, opened 2024-10-24
Remove invalid version in admcheck PyPI package

mgdcvetko:
We noticed that an invalid version made its way into one of our advisories (sorry) and we fixed it in our report with the last push of new advisories. This PR would fix it in OSSF advisory - valid version `0.0.9` is already listed so this is all that is needed to fix the issue.


rhalar (commented):
@calebbrown  We have multiple entries for the same advisory (`RLMA-2024-03512`) in `malicious-packages-origins`, presumably because the hash changed? Should we remove the first entry with the non-matching hash as well?

We modified the existing advisory since you were reprocessing the whole bucket, seemed like the better choice; the same issue also applies to the erroneous NPM namespace advisories because we fixed those manually as well.

calebbrown:
I inspected 0.0.9 and 1.0.0 of admcheck and they appear to be similarly malicious.

I think leaving this version in would make sense.

Also, we don't need to clean up the old hashes. That's a lot of clean up, and keeping them around doesn't hurt, and helps preserve the history of the repository.

rl-threat-detection:
The removed version has a trailing space, the correct `0.0.9` is in as well.

rhalar:
The removed version has a trailing space, the legitimate `0.0.9` is present as well.

calebbrown (approved):

calebbrown:
Thanks @rhalar I hadn't noticed the space!

PR #701, opened 2024-12-09
Fix for several invalid versions

mgdcvetko:
Hi, we noticed that five packages had invalid versions listed in their advisories. It's the same mistake across all advisories - version `1` is listed, while malicious packages actually published version `1.0` and never just `1`. 
This PR would fix this in all five OSSF published advisories, and I will also fix this in our reports in the AWS bucket.

Mentioning @rhalar for visibility.


osv/malicious/pypi/aiotrans/MAL-2024-9938.json
```diff
@@ -15,7 +15,7 @@
       "versions": [
         "1.0.0rc1",
         "1.0.0rc2",
-        "1",
+        "1.0",
```
rhalar, line 18: I have to bring this up again, just to make sure we're in sync with @calebbrown, and handling this correctly.

So _our_ procedure usually is that when non-additive changes happen to an advisory, we withdraw the advisory and generate a new one, with a new ID. This generates 'extra' advisories and withdrawals, and since the OSSF doesn't handle withdrawals automatically we decided to fix up existing ones manually for these smaller scale issues. This was already done in https://github.com/ossf/malicious-packages/pull/673

This is why there are two entries in `malicious-packages-origins` for `RLMA-2024-03512`
https://github.com/ossf/malicious-packages/blob/main/osv/malicious/pypi/admcheck/MAL-2024-4732.json#L48

@calebbrown commented that this is okay, since the 'history' is preserved https://github.com/ossf/malicious-packages/pull/673#issuecomment-2451091193

But that worked because the whole bucket was reprocessed from the start last time, which I don't believe will be the case anymore?
If we plan to manually edit existing advisories now, we probably need to generate a new entry in `malicious-packages-origins` as well?

calebbrown: If an additional version is *added*, it might be better to publish a new updated version of the report with the correct version(s), with a higher number (or an RLUP- report).

This would be merged and add the version the report.

*Removing* the `"1"` currently must be done manually.

If you'd prefer I can update all the versions in this PR.

calebbrown:
This PR is not entirely necessary as far as Python is concerned.

According to [the version spec](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers) "1" and "1.0" refer to the exact same version of a package.

rhalar:
Hey, I've been away here and there and haven't been able to comment. In the end, we decided we'll just close this PR since the versions are technically correct.
We'll see how we handle these in the future, we are considering normalizing versions before pushing to OSV.

PR #702, opened 2024-12-09
Validate reports against the OSV schema.

calebbrown:
This ensures that all the reports in this repository conform to the schema.


calebbrown:
cc @andrewpollock

jessmcclintock (approved):

PR #714, opened 2024-12-29
Add report for unisocks npm package

6mile:
Added a report for the unisocks NPM package identified by GitHax.


calebbrown (approved):
Thanks for the report

PR #759, opened 2025-01-15
Added report for venmo-ui npm package

6mile:
Added report for venmo-ui npm package


calebbrown:
Please ignore the validate OSV check failing. I added a check against the schema, but we don't require the ID to be set which is causing issues.

calebbrown (approved):

PR #791, opened 2025-01-31
Adding new malicious npm packages

awsactran:
Adding new malicious npm packages


osv/malicious/npm/prajwal_kewat/MAL-0000-prajwal_kewat.json
```diff
@@ -0,0 +1,23 @@
+{
+    "modified": "2025-01-31T03:01:05Z",
+    "published": "2025-01-31T03:01:05Z",
+    "schema_version": "1.5.0",
+    "summary": "Malicious code in prajwal_kewat (npm)",
+    "details": "The package communicates with a domain associated with malicious activity.",
+    "affected": [{
+        "package": {
+            "ecosystem": "npm",
+            "name": "prajwal_kewat"
```
calebbrown, line 10: Is this a malicious package? It looks empty to me.

calebbrown (changes requested):
Hi, @awsactran the PR looks great. Just one package does not appear to be malicious - if it is not malicious can you remove it please?

Thanks!

osv/malicious/npm/prajwal_kewat/MAL-0000-prajwal_kewat.json
awsactran: Hi @calebbrown . Thanks so much for catching this. This was my bad. While looking into another similar package with the same prefix, I mistakenly created an entry for this. Thanks for calling this out. I have pushed a new commit to remove it. Thank you so much for your time.

calebbrown (approved):

PR #845, opened 2025-03-20
Add black spammer packages

behnazh-w:
This pull request adds 17 malicious PyPI packages, all of which are associated with the black spammer toolkit. These packages have been confirmed and removed by the PyPI security team.


calebbrown (approved):
Thanks for the reports!

This suite of packages appears like a hacking tool, which are not usually considered apart of this dataset.

However the use of obfuscation, and the many packages with similar names indicates that it is not attempting to present itself as a legitimate package - so I am happy to include the reports.

PR #849, opened 2025-03-28
Report ok-script pypi package

OmkarPh:
### 📁 Suspicious Binary Files

- 🚩 **Disguised Binary Executable** detected:
    - `ok/__init__.cp312-win_amd64.pyd` (2.7MB)
    - File classified as `application/vnd.microsoft.portable-executable` with extension `.exe`
    - Loaded at package import time via the `__init__.py` file

### 🔄 Dangerous Auto-Update Mechanism

- 🚩 The package contains a sophisticated auto-update system (`GitUpdater.py`) that:
    - Downloads code from remote repositories autonomously
    - Executes code in new virtual environments
    - Can kill processes and modify system PATH values
    - Manipulates file ownership using Windows APIs

### Imported System Modules

The package imports numerous sensitive modules:

- `psutil` (process manipulation)
- `win32api`, `win32security` (Windows system API access)
- `subprocess` (command execution)

### Auto-Update Exploitation Path

1. Auto-update system (`GitUpdater.py`) detects version differences
2. Downloads code from remotely controlled repository
3. Creates isolated Python environment
4. Executes the downloaded code with system privileges

```python
# Code from GitUpdater.py showing remote execution capabilities
def start_app(self):
    communicate.update_running.emit(True, True)
    logger.info(f'start_app enter')
    try:
        new_ver = self.starting_version
        entry = 'main.py'

        script_path = os.path.join('repo', new_ver, entry)
        # [...snip...]
        process = subprocess.Popen(
            [python_path, script_path, f'--parent_pid={os.getpid()}'],
            creationflags=subprocess.CREATE_NO_WINDOW,
            close_fds=True,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE
        )

```


calebbrown:
Hi, thanks for your report.

I am looking at the "ok-script" project and am having trouble understanding how this is malicious.

There are definitely aspects of the code that are unusual, and potentially problematic, but I can't see anything that is directly malicious (I am inspecting both versions prior to 0.0.514 and afterwards).

The `.pyd` file is unusual in a .tgz, but may just be an artifact of packaging.

Are you able to provide more detail about why 'ok-script' is considered malicious?


Thanks!

abhisek:
@calebbrown Good point. We too had similar doubts initially but given strong signals decided to report it. Specifically

1. The associated GH repo has no code https://github.com/ok-oldking/ok-script
2. There was `GitUpdater.py` in `0.0.504` which seems to have been compiled to `pyd` in `0.0.507` as `
ok/update/GitUpdater.cp312-win_amd64.pyd`. This has pretty much download-exec capabilities and killing processes (Windows)
3. `0.0.504` also had `ok/update/GithubMultiDownloader.py` which had a list of proxy URLs. Guessing they are to bypass potential country level access blocks

Based on these strong indicators we decided to report, especially since any of the external URLs used for update process can, in future, be used to perform a download-exec on systems where this package is installed.

Do you mind keeping this open for a while even if these indicators do not confirm malicious behaviour? Let me do a deep dive on the code and what is changing across versions before concluding?

PR #859, opened 2025-04-28
Validate source IDs conform to `[a-z0-9-]+` and fix oracle reports.

calebbrown:
Ensure the sourceIDs that are encountered are valid when reports are parsed.

This also includes ensuring that the details section can be parsed as well.

"Oracle using Macaron" records now use the source "oracle-using-macaron" to conform.

### DIFF
--- a/internal/report/details_test.go
+++ b/internal/report/details_test.go
@@ -13,24 +13,35 @@
 // limitations under the License.
 
 //nolint:goconst
-package report_test
+package report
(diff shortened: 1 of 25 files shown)


calebbrown:
@behnazh-w - I have updated the reports you've submitted. Sorry this wasn't picked up earlier!

distractible (approved):

PR #880, opened 2025-05-05
Adding new malicious package entry for NPM package @moxfive-llc/common

awsactran:
Adds OSV entry for `@moxfive-llc/common` (`MAL-0000-moxfive-llc-common`).

Malicious filepath: `package/src/utils/has-global-action.ts`


calebbrown:
Hi Chi,

Thanks for the report.

I do not think this package is malicious.

The code only hits the URL when the `hasGlobalAction` function is called, but otherwise does nothing else.

Caleb

awsactran:
@calebbrown ,

Thank you for catching this. This was my bad as when I looked at the index.js I thought it calls `hasGlobalAction`. Closing this PR

PR #909, opened 2025-05-05
Add malicious package entries

awsactran:
This PR aggregates OSV entries for newly discovered malicious NPM packages.


osv/malicious/npm/epic-ue-saleforce-form/MAL-0000-epic-ue-saleforce-form.json
```diff
@@ -0,0 +1,34 @@
...
+  "modified": "2025-05-05T21:14:46.113Z",
+  "published": "2025-05-05T21:14:46.113Z",
+  "schema_version": "1.5.0",
+  "summary": "Malicious code in epic-ue-saleforce-form package (npm)",
+  "details": "The package communicates with a domain associated with malicious activity.",
+  "affected": [
+    {
+      "package": {
+        "ecosystem": "npm",
+        "name": "epic-ue-saleforce-form"
```
calebbrown, line 11: Hi Chi,

`epic-ue-saleforce-form` does not appear to be malicious.

It is merely logging the result from querying the `oastify[.]com` domain on import.

If you can remove this report, that would be great.

calebbrown (changes requested):
Hi Chi,

Thanks for all the reports. They all look fine, except for one which I've left a comment on specifically.

Caleb

osv/malicious/npm/epic-ue-saleforce-form/MAL-0000-epic-ue-saleforce-form.json
awsactran: Hi @calebbrown ,

Thank you for taking the time to review my commit.

I completely understand your assessment, and I respect your decision. That said, I was wondering if you'd be open to reconsidering the classification, or at least the broader implications of such behavior.

The reason I flagged the package is that, from a typical JS user's perspective, simply installing it triggers an automatic HTTP request to an external server without user awareness or consent. While I agree it's not explicitly exfiltrating sensitive data, the fact that it performs a network callback by default could be seen as suspicious IMO. 

For example, in some environments, automatic HTTP headers (including cookies or authorization headers) might unintentionally reveal sensitive metadata.

I fully recognize that intent and context matter, and of course, your decision is final. I just wanted to surface this concern in case it might be worth re-evaluating from a user safety and transparency standpoint. If you still feel it doesn't meet the criteria, I’ll be happy to close or retract the submission.

Thanks again for maintaining the OSSF Malicious Packages repo and your ongoing work to keep the ecosystem safer. . Thank you so much, Caleb!

calebbrown: Hi Chi,

I appreciate the feedback, and understand there is a lot of nuance.

I have not dug into what automatic HTTP headers are added by JavaScript frameworks like axios, and that is certainly an avenue for exfiltrating data without it appearing obvious.

In the case of `epic-ue-saleforce-form` the request is only triggered on import, so a user would have to call `require('epic-ue-saleforce-form')` from their code for it to be triggered.

Since this is not an "on install" triggered request, requires explicit usage, and it merely calls the endpoint without any additional data (even headers), I don't think it meets the bar of malicious as defined in the README.

That said, I do think it is a nuisance package that no one should be using. But I don't think it belongs in this dataset.

Caleb

awsactran: Hi Caleb, 

Thank you so much for your review and due diligence. I have now removed the entry. The PR has been updated for your review again.

awsactran:
@calebbrown 

Thank you so much for your review and due diligence. I have now removed the entry. The PR has been updated for your review again.

calebbrown (approved):

PR #930, opened 2025-06-12
Remove the overly broad match from GHSA to limit scanner noise.

calebbrown:
The GitHub repo github.com/primefaces/primereact-sass-theme uses a package.json file for this name squatting attack and has a version 10.8.5.

The overly broad version range from GitHub that covers every version means that this repo is matched and if someone has the repo checked out it will be potentially considered malicious by a scanning tool.

This change leaves the original version range from @awsactran.

### DIFF
--- a/osv/malicious/npm/primereact-sass-theme/MAL-2025-4574.json
+++ b/osv/malicious/npm/primereact-sass-theme/MAL-2025-4574.json
@@ -22,14 +22,6 @@
               "introduced": "100.0.0"
             }
           ]
-        },
-        {
-          "type": "SEMVER",
-          "events": [
-            {
-              "introduced": "0"
-            }
-          ]
         }
       ],
       "database_specific": {


elitsa-gosst (approved):

0xfatty:
This behavior seems to be happening for awhile from Github, and some other IIRC. Do you think this will need a quick campaign to address this in the repo? 

According to some conversations I have seen, the main reason they did this probably because of an assumption that once a package turned malicious, none of any other versions should be trusted.

awsactran:
This behavior seems to be happening for awhile from Github, and some other IIRC. Do you think this will need a quick campaign to address this in the repo?

According to some conversations I have seen, the main reason they did this probably because of an assumption that once a package turned malicious, none of any other versions should be trusted.

My quick script shows there seems to be at least hundreds of packages being tagged with version 0.

PR #945, opened 2025-07-21
Added report for 5 packages that were recently attacked

6mile:
Added specific versions of existing legitimate packages that were part of recent software supply chain attack.

Packages affected:
eslint-config-prettier: versions 8.10.1, 9.1.1, 10.1.6, 10.1.7
eslint-plugin-prettier: versions 4.2.2, 4.2.3
synckit: version 0.11.9
@pkgr/core: version 0.2.8
napi-postinstall: version 0.3.1

References for the attack:
https://www.bleepingcomputer.com/news/security/popular-npm-linter-packages-hijacked-via-phishing-to-drop-malware/
https://socket.dev/blog/npm-phishing-email-targets-developers-with-typosquatted-domain
https://socket.dev/blog/npm-phishing-campaign-leads-to-prettier-tooling-packages-compromise


calebbrown (changes requested):
Thanks for creating these!

Can you please add the blog references in the PR as references in the OSV reports.

i.e:
```
"references": [
  {
    "type": "WEB",
    "url": "..."
  }
]
```

It may also be worth linking to the issues in GitHub and the GHSA as well.

PR #960, opened 2025-08-08
Migrate from osv-scanner/pkg/models to osv-schema/bindings/go

calebbrown:
The models have moved. This updates Malicious Packages to point to the new, permanent location.

Unfortunately PURLToPackage didn't make it, so this has been forked.


calebbrown:
cc @another-rex

internal/report/validate.go
```diff
@@ -149,3 +166,70 @@ func validatePURL(ecosystem models.Ecosystem, name, purl string) error {
...
+	if !ok {
+		return osvschema.Ecosystem("")
+	}
+
+	wildcardRes, hasWildcard := ecoMap["*"]
+	if hasWildcard {
+		return wildcardRes
+	}
+
+	ecosystem, ok := ecoMap[pkgURL.Namespace]
```
distractible, line 201: I guess this is to match previous behaviour, but I guess I'd expect a more specific namespace to override *

calebbrown: That's true. Currently wildcard "*" entries are mutually exclusive to those with non-wildcard namespaces, so this isn't an issue.

But I have refactored it anyway to make the behaviour more intuitive.

Issue #910, opened 2025-05-08
Create statistics for monthly submissions by ecosystem


6mile:
I brought this up today with @jeffmendoza and @calebbrown in the Securing Critical Projects Working Group:  
I would love to see automatically generated statistics for several things:

1. total malicious package submissions each month
2. total malicious package submissions by language ecosystem (npm, pypi, etc)
3. total malicious package submissions per year
4. total malicious package submissions by language ecosystem per year
5. total malicious packages 
6. total malicious packages by language ecosystem

I am compiling some of these numbers myself, but it's painful as I do it based on the GitHub repo files in https://github.com/ossf/malicious-packages/tree/main/osv/malicious

PR #974, opened 2025-09-03
Add support for changing how counts are grouped.

calebbrown:
This allows:
- grouping by ecosystem to be enabled/disabled.
- grouping by month, year or all time.

To support this a simple helper library called "collector" was added.

#910


6mile (approved):
Looks good

PR #991, opened 2025-09-15
Update ingest workflow comment and schedule

conorfitch:
I've updated the comment beside the schedule in `ingest-cloud.yml` to correctly reflect that it's set to occur every half-hour, instead of every hour.

On a bigger matter, I have recently been wondering why I could not see some widely-reported malware advisories in the OSV dataset. I realised this was down to the workflow to ingest GHSA malware is set to occur every 24 hours, so it could be up to a day before some critical malware alerts make it from there into here.

This PR suggests a change of this schedule to occur every half-hour, to be in line with the schedule of ingesting OSV cloud data, for those who could look at this dataset to be aware of high-impact malware and need to act quickly.

Thank you for all your contributions to this project, it is fantastic!


calebbrown (changes requested):
Thanks for the suggestions.

Once a day for GHSA is not very frequent either.

May I also suggest updating the snippet of Python below to reduce the "TIMESINCE" value from 48h to 24h so we are not pulling as many entries.

calebbrown:
Thanks for making these changes. I recently noticed that the GHSA ingestor appears to be creating different hashes for the same content. This causes lots of noise for withdrawn packages.
I will pause on this PR until I have investigated that issue to avoid creating needless noise.

calebbrown:
#1002 is tracking this work. I have found the cause and solution for the bug, but need to have some upstream repositories updated.

Once resolved, I should be able to merge this PR.

calebbrown (approved):

PR #997, opened 2025-09-16
Manually run "Ingest GHSA Malware" & Assign IDs

Newbie012:
The current reports are out of date and mark the entire libraries as malicious even though only a subset of versions of them is.


calebbrown:
We have manually dealt with this separately. I have removed the `"introduced": "0"` as well.

PR #1031, opened 2025-10-02
Moved to correct location malicious package report 

socketz:
Moved to correct location malicious package report for @igh/server-updater-ssh-l1


calebbrown:
Thanks for your PR. I just merged #1032 that did the same thing.

Issue #858, opened 2025-04-23
Integrating malicious PyPI packages source


kam193:
Hey 👋 

I'd like to discuss integrating [my collection](https://github.com/kam193/package-campaigns) of malicious PyPI packages with the OSSF dataset - especially if it makes sense to think about it.

The dataset is based on my private hunting and I do it already for more than a year - with the automated publishing to the repo since summer last year. All published reports are manually verified. I collect clearly malicious as well as more unclear cases (categorized; my idea would be to go with malicious and eventually think about "probably pentest" aka potentially research or pentest activities with low harm).

Besides the question if you're willing to accept data from a private person (fully understand if not), I see so far following potential issues:

* data not in OSV format - this is easy to change,
* no versions tracking - I'm targeting packages intended for malicious purposes, not hijacked versions, so it wasn't a need so far. I do think about including versions in the future, but it may create additional work for me to verify each version.
* descriptions are organized over campaigns, including 1 or more packages. It means that sometimes the description may not exactly fit each package if it belongs to a campaign that is split over multiple packages or evolves (examples: [[1]](https://bad-packages.kam193.eu/pypi/campaign/2025-02-alicloud-client/), [[2]](https://bad-packages.kam193.eu/pypi/campaign/2025-04-zscaner/))
* I'm not always clearly marking if the malicious action is on installation, importing or calling a specific function. 
* Some campaigns can be too generic (e.g. https://bad-packages.kam193.eu/pypi/campaign/GENERIC-standard-pypi-install-pentest/)
* There is currently no S3 with data to download, just the public repository.

Please let me know what you think, if it makes sense to work on integrating my collection into OSSF dataset, and if you see other issues :)

calebbrown:
Hi @kam193,

Sorry for the slow reply. I would be happy to work with you on ingesting your data automatically. PyPI related coverage is an area I am keen to expand on.

One option for ingesting your data would be to build a tool that transforms your data into a series of OSV formatted files that can be ingested - a bit like how GHSA data is ingested.

I could also adapt the cloud ingestor to be able to target Git-based datasets as well.

Caleb

kam193:
I've added exports to OSV format - would be great to review it, please let me know if you had any comments. In the "credits" part, I've added myself as reporter if I have tracks of sending the report, or as analyst otherwise (usually when I flag a package that has already been removed or quarantined).

calebbrown:
Hi @kam193, I have nearly prepared the code on my side for integrating.

[...]

PR #1067, opened 2025-11-11
Add support for OSV reports ingesting from Git sources.

calebbrown:
This enables ingestion from sources hosted in GitHub, such as #858.

The config can be updated to look like the following:

```yaml
sources:
- id: "git-based-source"
  storage:
    type: "git"
    bucket: "https://example.com/repo.git"
    branch: "osv-branch"
  prefixes:
  - "path/to/reports/"

### DIFF
--- a/internal/sourceio/gitstorage.go
+++ b/internal/sourceio/gitstorage.go
@@ -51,7 +51,7 @@ func (s *GitStorage) Walk(ctx context.Context, prefix, start string, walkFn Walk
 		return "", fmt.Errorf("no repository specified")
 	}
 	if s.Branch == "" {
-		s.Branch = defaultGitBranch
+		return "", fmt.Errorf("no branch specified")
 	}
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: internal/sourceio/gitstorage.go [hunk 1]
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// StorageType implements the Storage interface.
+func (s *GitStorage) StorageType() StorageType {
+	return StorageTypeGit
+}
+
+// String implements the Storage interface.
+func (s *GitStorage) String() string {
+	return fmt.Sprintf("git: '%s - %s'", s.Repository, s.Branch)
```
distractible: It's perhaps a bit weird that String() would return different things after calling Walk (if s.Branch was not set)

calebbrown: Fixed. PTAL.

I now apply the default at parse time rather than in Walk.

PR #1091, opened 2025-12-02
Add github.com/kam193/package-campaigns as a source.

### LINKED ISSUE #858: Integrating malicious PyPI packages source

calebbrown:
This fixes #858.


elitsa-gosst (approved):

kam193:
Hey, thanks, it's perfectly fine! 

Thanks for the implementation and I want to say I'm always open to any feedback - so if you find anything I should fix or can improve in my reporting, please let me know anytime :)

PR #1093, opened 2025-12-02
Add report for the gtkfuscator PyPI package

behnazh-w:
Add report for the gtkfuscator PyPI package.


kam193:
Nice, I might have been too tired when I was looking at it to notice all parts

PR #1094, opened 2025-12-02
Unblock assign ids due to a name case problem for PyPI.

calebbrown:


kam193:
The context is that Python package names have normalization rules, and according to them, both names are equal: https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization

Issue #1104, opened 2025-12-10
False positive anaconda-anon-usage


jezdez:
This was reported in https://github.com/anaconda/anaconda-anon-usage/issues/199 and https://github.com/anaconda/anaconda-anon-usage/issues/201, to quote:

> I assume AWS Inspector has imported the finding from OSSF (@ossf) into their database recently, as this vulnerability only started to be reported on Wednesday 3rd December for our AWS accounts.

Probably when https://github.com/ossf/malicious-packages/issues/858 was merged.

`anaconda-anon-usage` is a package distributed with the commercial [Anaconda Distribution](https://www.anaconda.com/download) and is not available on PyPI, and never was: https://pypi.org/project/anaconda-anon-usage/.

This is a false positive; please remove it as soon as possible since Anaconda customers are falsely seeing reports in their use of AWS.

Other links:

- https://github.com/kam193/package-campaigns/pull/3
- https://github.com/kam193/package-campaigns/issues/4
- https://www.anaconda.com/blog/changes-to-anacondas-anonymous-usage-data-collection
- https://www.anaconda.com/blog/introducing-enhanced-usage-insights-for-anaconda-packages
- https://www.anaconda.com/download

kam193:
I'm updating the report to contain the version but please see https://github.com/kam193/package-campaigns/issues/4#issuecomment-3638893487 for explanation - the report is about a malicious package that was in PyPI, and not the one from Anaconda ecosystem. The report is intended for the PyPI ecosystem. I'm sorry for the confusion

Issue #1099, opened 2025-12-08
Remove "introduced: 0" from confluent-docker-utils


kam193:
As per https://github.com/kam193/package-campaigns/issues/2 it's turned out some tools have confused the non-PyPI package with the one from PyPI, and started flagging a safe package as malicious. To assist with that, I have updated my old report to include the specific package version instead of a generic "introduced: 0" event. Unfortunately, the merged report did not reflect the change:

https://github.com/ossf/malicious-packages/blob/cb0f9701a7e5e6c2fe30b51147dc3b105f06132d/osv/malicious/pypi/confluent-docker-utils/MAL-2024-9963.json

I'm not sure if it's because of older report versions present in the `malicious-packages-origins` or because it does not automatically remove data from the `affected` key and it should be done manually. @calebbrown could you please assist with the correct solution?

calebbrown:
Sure! Unfortunately the merging is additive, this means that correcting entries has to be done manually.

This just involves manually modifying the MAL-2024-9963 report.

PR #1108, opened 2025-12-11
Remove open ranges from reports introduces in kam193's PyPI collection

kam193:
Issues #1104 and #1099 showed that tooling using the database is most probably not able to distinguish between PyPI and non-PyPI packages. While I believe it's an important thing to develop further in the tooling*, causing unnecessary confusion is counterproductive for security. As such, I've populated my older reports with versions I'm aware of existing in PyPI in the past, and this PR aims to clean up open ranges from final reports, where they were introduced by my reports.

*I now have a great idea for a security-tool-confusion attack :smile:

### DIFF
--- a/config/start-keys.yaml
+++ b/config/start-keys.yaml
@@ -1,9 +1,9 @@
 amazon-inspector:
     IN-MAL-: IN-MAL-2025-149364.json
 kam193:
-    pypi/packages/malicious/osv/: a5c082ee4dcd70f9a1e898d5d6a4fa2ec4351406
+    pypi/packages/malicious/osv/: b1f700d92062ab1b16286da3db1111b830ea1d09
     pypi/packages/pentest/osv/: 0d65fa30569acb74a4cd2f6968297f9cf794b510
-    pypi/packages/probably_pentest/osv/: 55e61006a36c4dc9ecee9c3cbe74f03e64909e2e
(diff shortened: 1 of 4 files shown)


kam193:
The change was done with the following Python script:

```python
import json
def fix_osv(path):
    with open(path, 'r') as f:
        data = json.load(f)

    if "ranges" not in data["affected"][0]:
        return
    
    if data["affected"][0]["ranges"] != [{"type": "ECOSYSTEM", "events": [{"introduced": "0"}]}]:
        return
    
    # verify the ranges come from my data
    original_reports = data["database_specific"]["malicious-packages-origins"]
    is_my_data = False
    for report in original_reports:
        if report["source"] == "kam193":
            is_my_data = True
            continue
        if report.get("ranges", []) != []:
            # someone else added ranges, not modifying
            print(f"Not modifying {path}, ranges added by {report['source']}")
            return
        
    if not is_my_data:
        return
    
    del data["affected"][0]["ranges"]
    with open(path, 'w') as f:
        json.dump(data, f, indent=2)
        f.write('\n')  

import pathlib

for filepath in pathlib.Path('osv/malicious/pypi').rglob('MAL-*.json'):
    fix_osv(filepath)
```

And after that also manually verified to revert any unnecessary changes, as apparently Python and Golang apply encoding in JSON differently. Anyway, second look would be much appreciated - only removal of range keys is expected.

kam193:
It should now be okay. I also haven't found any more reports with such an issue

Issue #430, opened 2024-04-05
Be more resilient to failures during preprocessing.


calebbrown:
When a report stop preprocessing all new reports are blocked from being added.

Failing reports should be moved out of the way so that new reports can continue to be ingested, with a manual step to fix and merge in the failed reports.

calebbrown:
This is mitigated by #1110.

#1111 ensures pull-requests are still validated.

The remaining work is to add a workflow that runs once a weekday and complains loudly if any reports are under `osv/unmergable`

PR #1110, opened 2025-12-15
Move aside unmergable reports during preprocessing.

calebbrown:
This change helps ensure that any issue that prevents a report from being merged does not stop other reports from being merged.

In the future I will add a workflow that will fail (on weekdays) if any reports are present in osv/unmergable.

Some notes:
- unmergable reports are included in dupe checks during ingestion, preventing re-processing new reports
- unmergable reports are included in validation checks
- a flag for aborting if unmergable reports exist has been added so the validation check can still be used


calebbrown:
Fixes #430

distractible (approved):

PR #1201, opened 2026-04-13
fix: remove dead true|| in beforeListFunc to restore blob resume

suhr25:
## SUMMARY

This PR fixes a bug in `internal/sourceio/blobstorage.go` where resume optimization for blob storage was unintentionally disabled. As a result, incremental ingests always scanned from the start instead of resuming from a given key.

---

## FIX

**Before:**

```go
if true || start == "" {
    return func(_ func(interface{}) bool) error { return nil }
}
```

**After:**

```go
if start == "" {
    return func(_ func(interface{}) bool) error { return nil }
}
```

---

## VERIFICATION

With a non-empty `start` key, listings now begin from that offset instead of the start of the bucket. This avoids fetching already-processed objects, reducing API usage and speeding up incremental runs, while existing safeguards still ensure correctness.


calebbrown (approved):
Great find!

PR #1227, opened 2026-04-30
added malicious packages from user saif7777

KunalSin9h:


osv/malicious/npm/bpmn-studio/MAL-0000-bpmn-studio.json
```diff
@@ -0,0 +1,36 @@
...
+      "ranges": [
+        {
+          "type": "SEMVER",
+          "events": [
+            {
+              "introduced": "0"
+            }
+          ]
+        }
+      ],
```
calebbrown: It might be worth removing this range. It appears as though `bpmn-studio` was published and removed previously.

KunalSin9h: @calebbrown deleted that file, since the package is removed completely.

calebbrown (approved):

PR #1229, opened 2026-04-30
fix(@cap-js): remove open-ended SEMVER ranges for db-service, postgres, sqlite

patricebender:
The current OSV entries for three `@cap-js` packages contain open-ended SEMVER ranges (`introduced: X` without a `last_affected` or `fixed`) that flag **all versions >= the compromised version** as malicious. This causes false positives for all users who have upgraded past the compromised version.

Clean fixed versions have since been published by the legitimate SAP maintainers.

Removed the `ranges` block from all three affected entries, keeping only the explicit `versions` list.


calebbrown:
I have merged #1230 which makes this change. Thanks for your pull request.

PR #1234, opened 2026-05-03
Linter version changed. Fix the warnings.

calebbrown:


kam193:
A small idea in the context: maybe we don't need to run go linter when only OSV reports are changed? Not a big impact, but always a few CPU cycles less :)

calebbrown:
> A small idea in the context: maybe we don't need to run go linter when only OSV reports are changed? Not a big impact, but always a few CPU cycles less :)

Good idea. Done!

distractible (approved):

PR #1262, opened 2026-05-19
added malicious PCP Microsoft/durabletask pypi package

KunalSin9h:
https://safedep.io/malicious-durabletask-pypi-supply-chain-attack/

Screenshot_2026-05-19_23-40-04


KunalSin9h:
some mess with git commits. re opening pr.

PR #1276, opened 2026-05-26
Withdraw FastAPI v0.136.3 and other FPs reports

awsactran:
Withdraw FastAPI v0.136.3 due to false positive. This package was flagged because of the consumption of indirect dependency fastar. After further deep dive, the dependency was developed by FastAPI team and the whole package is in a good shape.

Issue #1274, opened 2026-05-26
FastAPI v0.136.3 is marked as malicious by AmazonInspect / OSV.


KunalSin9h:
The description seems to be a speculative assumption, not concrete evidence. This is False Positive to us, and as SafeDep consumes OSV, is might be issue with downstream consumers of SafeDep services. 

https://osv.dev/vulnerability/MAL-2026-4750

This is the PR which added `fastar` dependency to FastAPI project. https://github.com/fastapi/fastapi/commit/4b264878d71b2dc0e9e0f90ef15459d4302a7f9c

KunalSin9h:
@calebbrown please look into it.

mmabramovv:
Looks pretty wrong tbh:
https://pypi.org/project/fastar/ - drop-in Rust wrapper for tar/flate2/zst archives
Exact commit where it got added: [4b26487](https://github.com/fastapi/fastapi/commit/4b264878d71b2dc0e9e0f90ef15459d4302a7f9c#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R62-L154)
The dev did lock deps changes: https://github.com/fastapi/fastapi/commit/622dcdc99c9561bacce7f292bc0b38f49ec8fd57

@tiangolo what do you think?

UPD: removed the name mention

joniumGit:
I think the `fastar` was previously pulled in transitively from `fastapi-cloud-cli` https://github.com/fastapilabs/fastapi-cloud-cli/commit/0499ba9922b85b6c73f8e4ad7db9000b6b713384. The commit https://github.com/fastapi/fastapi-cli/commit/7f4e88776692f7f05dce64106c6f202eb0097eb4 (Jul 7, 2025) added the cloud cli dependency and the standard option was introduced to `fastapi` already in https://github.com/fastapi/fastapi/commit/a25c92ceb9ab5be49901443d7184c149797192df (Aug 2, 2024).

This looks like a false positive due to a transitive dependency being forced to a newer version in fastapi and not because of a malicious change. Also, if you look at the commit in Github the change is present in tags from 1.136.0 in `fastapi`, this alert is missing most of them.

calebbrown:
@awsactran mind having a look?

tiangolo:
I own and maintain FastAPI, this is a false positive.

[...]

PR #1278, opened 2026-05-27
Unwithdraw reports accidentally withdrawn in #1276

calebbrown:
#1276 withdrew some FP reports, but it also withdrew some real reports.

This change reinstates the real reports.

The checks were based on:
- packages not being available in the registry
- additional sources being present


calebbrown:
Issue: #1274

elitsa-gosst (approved):

PR #1289, opened 2026-06-02
During Merge, ensure Normalize happens before name check.

calebbrown:
Some of the unmergable reports are because a report with a non-canonical name was attempting to be merged with one with a canonical name.

Before this change the name check happened before Normalize() was called on the "other" report. This meant that the package name had not yet been canonicalized.

This change reverses this behaviour. So the names will be canonicalized before the name check and avoid some merges from failing.


internal/report/merge_test.go
```diff
@@ -34,6 +34,14 @@ func TestMerge_MismatchName(t *testing.T) {
 	}
 }
 
+func TestMerge_CanonicalizeName(t *testing.T) {
+	r := testReport(osvschema.EcosystemPyPI, "this-is-a-package")
+	other := testReport(osvschema.EcosystemPyPI, "this.IS-a_package")
+	if err := r.Merge(other); err != nil {
+		t.Fatalf("Merge() = %v; want no error", err)
+	}
```
distractible, line 42: it'd be good to have the converse as well

distractible (approved):

internal/report/merge_test.go
calebbrown: `dest.Merge(other)` expects `dest` to already be normalized, so I'm not sure a converse test makes sense.

PR #1293, opened 2026-06-04
Fix some name canonicalization quirks.

calebbrown:
Ensure report.Name is always canonicalized. This has the following impacts:
- ensures that name equality checks in Merge() work as expected
- ensures that two reports for the same canonical name, but with different un-canonical names land in the same path

The canonical name is not set in affected[0].package.name for the time being. This may be done in a future step.

This change allows some unmergable reports to be correctly merged.


kam193:
Have you verified the recent Amazon Inspector reports? At least the whatsfly_labfox looks like a potential FP related to their recent issues. I'd say it is dangerous, but I don't see clear proof of malicious behavior, and VT over the DLL is [inconclusive](https://www.virustotal.com/gui/file/335e20ee2aade3f96b825a49c30bfc24b4ce8607db3d1ccfe96285f35b42f1da/detection).

calebbrown:
> Have you verified the recent Amazon Inspector reports? At least the whatsfly_labfox looks like a potential FP related to their recent issues. I'd say it is dangerous, but I don't see clear proof of malicious behavior, and VT over the DLL is [inconclusive](https://www.virustotal.com/gui/file/335e20ee2aade3f96b825a49c30bfc24b4ce8607db3d1ccfe96285f35b42f1da/detection).

Thanks for the heads up, I'll check. This was added when they had the issue the other week.

kam193:
After looking at them, I have strong doubts only about whatsfly_labfox, the rest looks rather correct to me. It seems to be using downloading as a fallback, and I didn't find anything clearly malicious. I'd not recommend using it, but I cannot confirm the malicious behavior.

distractible (approved):

calebbrown:
@kam193 I will withdraw whatsfly_labfox after this PR is merged.

PR #1295, opened 2026-06-05
Add more IronWorm related packages.

calebbrown:
"crypto-javascri" appears to be the proving grounds of the malware payload.
"crypto-javascript" is the key connecting package between the earlier malicious packages and the active "IronWorm" compromises.


calebbrown:
Let me adjust the detail on these. These are malicious packages and not compromised.

calebbrown:
> Let me adjust the detail on these. These are malicious packages and not compromised.

Done

elitsa-gosst (approved):

PR #1305, opened 2026-06-10
Add missing versions for bramin and magique-ai malicious packages.

calebbrown:


calebbrown:
cc @kam193

kam193:
Good catch, I can confirm I've missed assigning these versions.

distractible (approved):

calebbrown:
versions were added in a source.

PR #1306, opened 2026-06-11
Add initial malosv command for managing the repo.

calebbrown:
Includes "withdraw" and "unwithdraw" subcommands, and a library for parsing report-based arguments.

The tool aims to make the management of the repository, and the creation of valid PRs easier.


internal/report/report.go
```diff
@@ -245,6 +245,24 @@ func (r *Report) IsWithdrawn() bool {
...
 
+// Withdraw marks the report as withdrawn at the given time.
+// It also updates the modified time to the same time.
+func (r *Report) Withdraw(t time.Time) {
+	r.raw.Withdrawn = t
+	r.raw.Modified = t
+}
+
+// Unwithdraw marks the report as unwithdrawn.
+func (r *Report) Unwithdraw() {
```
distractible, line 300: Presumably we don't want unwithdraw_at?

cmd/malosv/internal/cmd/withdraw.go
```diff
@@ -0,0 +1,88 @@
...
+	// Create a temp directory for atomic writes. We use the system temp
+	// directory to avoid accidentally leaving temp junk in the repository.
+	tempDir, err := os.MkdirTemp("", "osv-withdraw-*")
+	if err != nil {
+		log.Fatalf("Failed creating temp dir: %v", err)
+	}
+	defer func() {
+		// Clean up temp directory
+		if err := os.RemoveAll(tempDir); err != nil {
+			log.Fatalf("Failed cleaning up temp dir: %v", err)
```
distractible, line 65: any other cleanup that might be needed before a Fatal? (and in unwithdraw)

cmd/malosv/internal/cmd/unwithdraw.go
```diff
@@ -0,0 +1,79 @@
...
+			return fmt.Errorf("too many reports for %q (%d)", path, len(pkgReports))
+		}
+		targets = append(targets, pkgReports...)
+	}
+
+	log.Printf("Found %d target(s) for unwithdrawal", len(targets))
+
+	// Create a temp directory for atomic writes. We use the system temp
+	// directory to avoid accidentally leaving temp junk in the repository.
+	tempDir, err := os.MkdirTemp("", "osv-unwithdraw-*")
```
distractible, line 49: This might not be on the same filesystem? Maybe something with renameio.TempDir to handle that, even if it might end up leaving junk :|

cmd/malosv/internal/cmd/withdraw.go
calebbrown: Nope. This is the only clean up.

cmd/malosv/internal/cmd/unwithdraw.go
calebbrown: renameio.TempDir is interesting. I currently don't use it or check for cross filesystem operations in other places os.MkdirTemp is used, so I will follow up on this in a future later PR.

internal/report/report.go
calebbrown: OSV doesn't specify an "unwithdraw_at" field.

distractible: Oh, just that Modified was also controlled by the cli arg. I don't think we really need that though.

cmd/malosv/internal/cmd/unwithdraw.go
distractible: OK!

distractible (approved):

PR #1307, opened 2026-06-11
Lower the default perms and limit when creds are included.

calebbrown:


distractible (approved):
Neat, I assume it works :)

PR #1316, opened 2026-06-16
MAL-2025-6903: remove affected ranges

kytta:
About ten months ago, names of two packages developed internally at Hornetsecurity were used to upload malicious packages to NPM: [`@hornetsecurity/angular-pew-pew`](https://osv.dev/vulnerability/MAL-2025-6890) and [`@hornetsecurity/angular-search-box`](https://osv.dev/vulnerability/MAL-2025-6903). They later were removed by NPM and are now security-held packages. The `@hornetsecurity` namespace is also security-held.

The "problem" that we are facing now is that our dependency scanners continue flagging `@hornetsecurity/angular-search-box` as malicious, probably due to the `affected[0].ranges` setting specifying an open-ended range rather than individual packages:

https://github.com/ossf/malicious-packages/blob/2b49713871900265940d1f9bd2cc13333fda3022/osv/malicious/npm/%40hornetsecurity/angular-search-box/MAL-2025-6903.json#L14-L23

`@hornetsecurity/angular-pew-pew` is different in that it only mentions the individual affected versions. I believe this is more correct, as the packages are now on hold, and no-one, including us, can publish new versions. Thus, I propose removing the `ranges` field.

Please let me know if such change is acceptable and if you have any concerns :)


calebbrown (approved):

calebbrown:
This looks great. Happy to merge.

Issue #1315, opened 2026-06-15
Fields unification, data deduplication and presentation in OSV


kam193:
I've noticed a few UX points we should address one day, probably in coordination with other projects. Leaving an issue before I forget about it.

1) `database_specific.ioc.*` vs. `affected.[].database_specific.indicators.*`

The first comes from https://github.com/ossf/malicious-packages/blob/main/docs/schema_additions.md#iocs-fields-under-development and it's what I use, the second form is used by Amazon Inspector. I think logically both make sense: I currently collect IoC per-campaign, while Amazon Inspector is able to assign them to a specific version. However, the osv.dev treats them a bit differently: the first is directly visible, and the second needs explicit unhiding:

https://osv.dev/vulnerability/MAL-2026-5769 vs. https://osv.dev/vulnerability/MAL-2026-5786
<img width="1388" height="616" alt="Image" src="https://github.com/user-attachments/assets/643cb6f5-a314-4d3e-8eac-49cc8c7cff37" />

<img width="1334" height="1594" alt="Image" src="https://github.com/user-attachments/assets/3ead2a4c-47d2-4520-a83c-01dc95936e5e" />

Notably, the OSV.dev representation does not show which version these fields are related to (but it's still in JSON).

2) OSSF/MAL does not remove duplicated dictionary entries in `affected.[].database_specific.*.*`, which ends up in things like https://github.com/ossf/malicious-packages/blob/478789ba71d36d5b823e29c894c8fe4fa244c063/osv/malicious/npm/%40solana-labs/ancor/MAL-2026-5786.json#L22-L54 

3) Each report generates a new `malicious-packages-origins` entry, which is by default presented on OSV.dev. It works well when we have 1-3 source reports, but it starts to be a bit counterproductive after that. From my side, every update to the campaign overview causes updates to older reports and a new entry; Amazon Inspector often provides data in a few separated reports merged later by OSSF/MAL. And we also have other sources. And new versions. And we end up in OSV.dev looking like in https://osv.dev/vulnerability/MAL-2026-5786

<img width="696" height="1538" alt="Image" src="https://github.com/user-attachments/assets/635c8084-f06e-4699-9ce7-2b0ca490419f" />

I think it hides the more important information like references and details in next section. It would be great to have `malicious-packages-origins` hidden by default or put in a different place - this would require asking OSV.dev team for help.

PR #1317, opened 2026-06-17
fix: recursively merge and deduplicate database_specific maps and arrays

jeffrey-theog06:
Context / Problem
--------------------
Addresses [Issue #1315]
(https://github.com/ossf/malicious-packages/issues/1315) 
(Point 2).

When merging reports (such as multiple AWS Inspector reports or updates), duplicate entries inside lists in nested database_specific fields (like cwes list of objects or nested lists in indicators) were being preserved, causing duplicated data in the final merged OSV JSON files.

Furthermore, if a nested object like database_specific.indicators existed in both reports, merging them would silently discard keys that already existed, rather than merging their contents recursively.

Solution
---------------
-> Refactored combineDatabaseSpecific in internal/report/merge.go to recursively traverse and merge maps, and append/deduplicate slices.
-> Added helper function deduplicateSlice utilizing reflect.DeepEqual to correctly and generically deduplicate arrays of any types (including complex structs/dictionaries like cwes).
-> Added helper function sanitizeAndDeduplicate to ensure initial reports are deduplicated.
-> Updated TestMerge_AffectedDatabaseSpecific to expect duplicate-free arrays.
-> Added new unit test cases covering recursive merges and nested object list deduplication (TestMerge_AffectedDatabaseSpecific_Recursive, TestMerge_AffectedDatabaseSpecific_DeduplicateObjects).

Verification
--------------
All unit tests in internal/report/ were successfully updated and extended to cover nested structure merging and deduplication.

### DIFF
--- a/internal/report/merge.go
+++ b/internal/report/merge.go
@@ -238,41 +238,116 @@ func mergeSlices[K comparable](ss ...[]K) []K {
 // The following rules are used to determine how the merge works:
 // - sources *must* have keys unique to that source.
 // - values must not be scalars. they must be arrays or objects.
-// - values that are arrays are concatenated together, duplicates will be preserved.
-// - values that are objects are merged, keys should be unique, but if they are the same the first key seen will retain its value.
+// - values that are arrays are concatenated together, duplicates will be removed.
+// - values that are objects are merged recursively, keys should be unique, but if they are the same the first key seen will retain its value.
 func combineDatabaseSpecific(objs ...map[string]any) map[string]any {
 	if len(objs) == 0 {
 		return nil
 	}
-	res := make(map[string]any)
+	var res map[string]any
 	for _, obj := range objs {
+		if obj == nil {
+			continue
+		}
+		if res == nil {
+			res = make(map[string]any)
+			for k, v := range obj {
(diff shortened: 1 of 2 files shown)


calebbrown:
Hi, I appreciate your interest in contributing.

However a decision about how to solve this problem has not been made.

Recursing through an arbitrary JSON object tree and collapsing duplicate entries without understanding the semantic context of those entries may have unintended consequences that need to be considered.

jeffrey-theog06:
Thank you for the feedback Mr @calebbrown ! That is a very valid architectural concern.

The intent behind this PR was to address the specific issue of duplicate CWE objects and indicators (like domains/IPs) that get generated during ingestion/merging. Currently, the implementation only recurses within the database_specific map, but you are absolutely right that a completely generic recursive merge might have unintended consequences if other custom fields need to preserve duplicates.

If you prefer a more conservative approach, we can easily modify this PR to:
1. Restrict the deduplication strictly to known fields (like cwes and the lists under indicators).
2. Or, we can pause this PR and wait until the team decides on the preferred design/strategy for handling these fields.

Let me know how you'd like to proceed, and I'd be happy to adapt the code to match the preferred approach!

PR #1336, opened 2026-07-01
Remove malformed alias

zhelinliao-oss:
Remove malformed aliases `MAL-0000-ghsa-malware-* in malicious-packages records below:
MAL-2024-12168
MAL-2024-12164
MAL-2025-47033
MAL-2025-47051
MAL-2025-47032
MAL-2025-47045
MAL-2025-47048
MAL-2025-47046
MAL-2025-47047
MAL-2025-47029
MAL-2025-47040


kam193:
The malformed alias is clear, but I'm not sure what is the point in replacing more detailed information with an open range? The open range is quite often a source of troubles in cases of dependency confusion, etc. I'd not recommend removing them.

PR #1344, opened 2026-07-04
Malicious package: npm/datefmt-helper (download-and-execute dropper)

akyroslabs:
Adds an OSV record for the npm package `datefmt-helper`, a download-and-execute
supply-chain dropper disguised as a date-formatting utility.
  
A `postinstall` hook runs `node postinstall.js` on install, which downloads a
second-stage payload from the hardcoded IP 115.190.124.243 and executes it via a
Node child process — arbitrary code execution on any machine or CI runner that 
installs the package (versions 1.0.0, 1.0.1). 
  
Detected by codelake Research from the live npm feed on 2026-07-02; not present in
OSV or GHSA at the time of reporting.
  
IOCs:
- Remote endpoint: 115.190.124.243
- sha256 (datefmt-helper-1.0.0.tgz): f2f58e73becce20f28c6f18700731f48c067826defe8279da2e24e4f69c25c72
  
Full technical write-up: https://research.codelake.dev/advisories/clr-2026-2990-datefmt-helper

### DIFF
--- a/osv/malicious/npm/datefmt-helper/MAL-0000-datefmt-helper.json
+++ b/osv/malicious/npm/datefmt-helper/MAL-0000-datefmt-helper.json
@@ -1,42 +1,43 @@
 {
-    "modified": "2026-07-04T00:00:00Z",
-    "published": "2026-07-04T00:00:00Z",
-    "schema_version": "1.6.0",
-    "affected": [
-      {
-        "package": {
-          "ecosystem": "npm",
-          "name": "datefmt-helper"
-        },
-        "versions": [
-          "1.0.0",
-          "1.0.1"
-        ]
-      }
-    ],
-    "details": "The npm package `datefmt-helper` is a supply-chain dropper disguised as a benign date-formatting utility (its `description` reads \"dates formatting utility with locale support\"). The package ships no source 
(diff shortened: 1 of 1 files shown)


osv/malicious/npm/datefmt-helper/MAL-0000-datefmt-helper.json
```diff
@@ -0,0 +1,42 @@
...
+      } 
+    ],
+    "references": [
+      {
+        "type": "ADVISORY",
+        "url": "https://research.codelake.dev/advisories/clr-2026-2990-datefmt-helper"
+      } 
+    ],
+    "database_specific": {
+      "iocs": [
```
kam193: The https://github.com/ossf/malicious-packages/blob/main/docs/schema_additions.md#iocs-fields-under-development suggests keeping it as an object with subkeys. I know it's not perfect, but would you mind keeping the object structure, with e.g. additional key for the hash? This way scripts prepared for the suggested schema won't break when getting a list in `database_specific.iocs`

akyroslabs: Done, thanks! Switched iocs to the object structure (ips + hashes).

calebbrown (approved):

PR #1346, opened 2026-07-05
Malicious package: npm/realagent (curl-to-bash download-execute dropper)

akyroslabs:
Adds an OSV record for the npm package `realagent`, a download-and-execute supply-chain dropper disguised as an "MCP Server" ("RealAgent MCP Server").

A `postinstall` hook runs `node install.js` on install, which uses `curl` to fetch a remote shell script from the hardcoded endpoint http://60.247.61.162:8083/install.sh and executes it via a Node child process piped to bash (execSync) — arbitrary code execution on any machine or CI runner that installs the package (version 1.1.0).

Detected by codelake Research from the live npm feed; not present in OSV or GHSA at the time of reporting.

IOCs:
- Remote endpoint: http://60.247.61.162:8083/install.sh (and /api/version/latest)
- sha256 (realagent-1.1.0.tgz): 043af19d6aee166ef5cc796eeca42e2508664f9a574ff1c86052850e8a653f49

Full write-up: https://research.codelake.dev/advisories/clr-2026-2992-realagent

### DIFF
--- a/osv/malicious/npm/realagent/MAL-0000-realagent.json
+++ b/osv/malicious/npm/realagent/MAL-0000-realagent.json
@@ -1,0 +1,45 @@
+{
+  "modified": "2026-07-05T00:00:00Z",
+  "published": "2026-07-05T00:00:00Z",
+  "schema_version": "1.6.0",
+  "affected": [
+    {
+      "package": {
+        "ecosystem": "npm",
+        "name": "realagent"
+      },
+      "versions": [
+        "1.1.0"
+      ]
+    }
+  ],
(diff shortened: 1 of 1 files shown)


akyroslabs:
Withdrawing this report after a deeper re-review — it does not hold up, and I'd rather retract a false positive than leave it in the queue.

On closer analysis, `realagent`'s `install.js` performs a **standard checksum-verified prebuilt-binary install** (the same pattern as esbuild/sharp/@swc): it fetches version metadata from the project's own site-server, downloads the platform archive, **verifies its SHA-256 against the server-provided checksum**, then extracts (`tar -xzf`/`unzip`) and `chmod`s it into `~/.realagent/bin/`.

The `curl -fsSL …/install.sh | bash` string that I read as the payload appears **only inside two `warn()` log messages** as a manual-fallback hint printed when the automatic download fails — it is **never executed** by the script. The hardcoded IP is the package's **own declared homepage/site-server**, not third-party infrastructure.

I could not substantiate the download-and-execute classification, so I'm closing this. Apologies for the noise, and thanks for the review bandwidth.

PR #1352, opened 2026-07-07
fix: remove open range from mistralai package

KunalSin9h:
this package should not have range introduced 0, only 3 versions were malware.

### DIFF
--- a/osv/malicious/npm/@mistralai/mistralai/MAL-2026-3432.json
+++ b/osv/malicious/npm/@mistralai/mistralai/MAL-2026-3432.json
@@ -101,16 +101,6 @@
         "import_time": "2026-05-12T01:08:51.454429056Z",
         "id": "GHSA-3q49-cfcf-g5fm",
         "modified_time": "2026-05-12T00:28:13Z",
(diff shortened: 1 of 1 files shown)


KunalSin9h:
cc @calebbrown

calebbrown:
Thanks for the PR.

However the range is only in the database_specific section. I don't think this change is required.

KunalSin9h:
@calebbrown but isn't its wrong of having "all version", we (SafeDep) consume it, and it made issues with marking all version of this package as malicious, due to this @0.

kam193:
Shouldn't you use only the root key for determining the affected versions? `malicious-packages-origins` acts as a revision history with previous reports, there are plenty of cases where the versions were corrected afterwards, leaving the original affected reports in this historical field. I think you should use the root field to select affected packages, and the origins only in the second step to match affected versions with original reports

KunalSin9h:
@kam193 maybe, this was not handled from our side, will fix thanks for feedback.

kam193:
You're welcome :)

PR #1366, opened 2026-07-10
Add malicious package details for polymarket-mcp-v2

skb1209:
Added a JSON file detailing the malicious behavior of the polymarket-mcp-v2 package, including its postinstall hook and data exfiltration methods.


osv/malicious/npm/polymarket-mcp-v2/MAL-0000-polymarket-mcp-v2.json
```diff
@@ -0,0 +1,91 @@
...
+          "package_integrity": [
+            {
+              "filename": "polymarket-mcp-v2-2.1.6.tgz",
+              "hashes": {
+                "sha1": "cb215c192332b668d419e78ceab38b0312fc5ccc",
+                "sha256": "97690f6e2d4ce315f19595b04cae3afdac0bfe9d98b0d3caeb1fd99eab133cde"
+              }
+            }
+          ]
+        }
```
calebbrown: This structure is unsupported by Malicious Packages, and is only listing files in the package - please remove.

osv/malicious/npm/polymarket-mcp-v2/MAL-0000-polymarket-mcp-v2.json
```diff
@@ -0,0 +1,91 @@
...
+      }
+    }
+  ],
+  "references": [
+    {
+      "type": "PACKAGE",
+      "url": "https://www.npmjs.com/package/polymarket-mcp-v2"
+    },
```
calebbrown: This reference to the package is redundant and will 404 after the package is removed. Please remove.

osv/malicious/npm/polymarket-mcp-v2/MAL-0000-polymarket-mcp-v2.json
```diff
@@ -0,0 +1,91 @@
...
...dividually with a short delay between uploads.\n\nIdentifiers and file names are lightly obfuscated with Unicode escape sequences and reversed string literals (for example, \"nosj.di\".split(\"\").reverse().join(\"\") for id.json) rather than a string-array or bytecode packer.\n\nThe C2 (170.205.31.203:3001, hosted on Advin Services LLC / AS206216) also serves a Next.js application at /dashboard titled \"ENV Bot - HYPE\", self-described in its page metadata as an API wrapper around a wallet-management Telegram bot. A development-mode error response from the same host disclosed its filesystem path, C:\\1\\anti-server\\bot-wallet-management-v1\\, indicating a Windows-hosted operator panel sitting behind the exfiltration endpoint.\n\nThe exported function names, the exact sudo/ufw command sequence, the /api/scan-patterns, /api/block-patterns, and /api/v1 endpoint names, and the id.json/config.toml/.env target list all match the levex-refa/lint-builder payload documented in the February 2026 compromise of the dev-protocol GitHub organization. This package appears to be a redeployment of that same toolkit against new, previously unreported C2 infrastructure rather than a new campaign.",
```
calebbrown: This is description is overly verbose.

Ideally it should give a clear summary indicating what the payload does, and the impact to the victim to facilitate with detection and remediation. Details about how the malware works are very optional, but a consumer of the report should not need to read the entire report to understand what is happening.

In this case I can see that the malicious payload is executed on install, is an info stealer - likely the levex-refa/lint-builder payload.

calebbrown (changes requested):
Thanks for the review. I have added some feedback on the report.

rbnfrhnr:
Ran into another deployment of this one while going through new npm packages last week.

`app-soda-layer@2.1.6`, published 27 July, author pulled it himself 2h25m later. Same `from_str_1`/`from_str_2`, same authorized_keys + `chown -R` + `ufw enable` + `ufw allow 22/tcp` sequence, same `/api/ssh-key`, `/api/scan-patterns`, `/api/block-patterns` on 3001. Also version 2.1.6, which feels like more than coincidence. Different C2 though, 95.216.118.146 (Hetzner).

One difference: yours obfuscates the identifiers with the reversed string trick, mine doesn't. Plain `"id.json"` sitting in the array.

I pulled the SSH key off the C2 while it was still up and wanted to compare it against whatever 170.205.31.203 was handing out, but both hosts are timing out for me now (mine was still answering this morning, so it looks like it just got taken down). If you happened to save that key it would be worth comparing. Same key on both would settle whether it's one operator rather than a shared tool.

Mine is #1405 if useful

skb1209:
> Ran into another deployment of this one while going through new npm packages last week.
> 
> `app-soda-layer@2.1.6`, published 27 July, author pulled it himself 2h25m later. Same `from_str_1`/`from_str_2`, same authorized_keys + `chown -R` + `ufw enable` + `ufw allow 22/tcp` sequence, same `/api/ssh-key`, `/api/scan-patterns`, `/api/block-patterns` on 3001. Also version 2.1.6, which feels like more than coincidence. Different C2 though, 95.216.118.146 (Hetzner).
> 
> One difference: yours obfuscates the identifiers with the reversed string trick, mine doesn't. Plain `"id.json"` sitting in the array.
> 
> I pulled the SSH key off the C2 while it was still up and wanted to compare it against whatever 170.205.31.203 was handing out, but both hosts are timing out for me now (mine was still answering this morning, so it looks like it just got taken down). If you happened to save that key it would be worth comparing. Same key on both would settle whether it's one operator rather than a shared tool.
> 
> Mine is #1405 if useful

Sorry for the slow reply. I checked my saved artifacts — I didn't retain the /api/ssh-key response from 170.205.31.203 at capture time, and the host is down now, so I can't do the direct key-to-key comparison. I don't want to claim a shared operator I can't back with the key.

What I can confirm is code-level overlap, which points to the same toolkit rather than proving the same operator:

identical exported from_str_1 / from_str_2
the same authorized_keys → chown -R <user>:<user> ~/.ssh → ufw enable → ufw allow 22/tcp sequence
the same /api/ssh-key, /api/scan-patterns, /api/block-patterns, /api/v1 endpoints on port 3001
matching id.json / config.toml / .env targeting
both published as version 2.1.6

One build difference: mine (polymarket-mcp-v2) obfuscates identifiers with the reversed-string trick ("nosj.di".split("").reverse().join("")), yours uses plain literals — different builds of the same levex-refa/lint-builder toolkit against separate C2s. If you still have your key from 95.216.118.146, posting its fingerprint would let a future capture settle the shared-operator question.

I run a screening pipeline over new npm publishes, so I come across these Polymarket-impersonation variants fairly regularly. Happy to keep comparing notes when either of us hits a new one — cross-referencing keys and endpoints across captures seems like the fastest way to pin down whether it's one operator or a shared toolkit.

calebbrown (approved):

PR #1378, opened 2026-07-15
Handle recursive report ingestion - GHSA appears to now include MAL

calebbrown:
Some reports from GHSA are from this repository.

This change adds support for handling recursive reports and cleans up the GHSA reports.

The three behaviours can be configured per-source:
- fail: abort if a recursive report is detected (default)
- ignore: skip the recursive report entirely
- allow: handle the report as normal (the previous behaviour)

The reports starting with `MAL-0000-ghsa-` that are recursive have also been removed in this PR.


elitsa-gosst (approved):

sachin-sandhu:
@calebbrown , thanks for taking care of this promptly.

PR #1368, opened 2026-07-12
Add malicious npm package polymarket-mcp-v2@2.1.6 (SSH-key backdoor)

akyroslabs:
Reporting a confirmed-malicious npm package identified by codelake Research from the live npm feed.

**Package:** `polymarket-mcp-v2@2.1.6` (npm) — typosquat of a Polymarket MCP server, empty description, 3-file shell.

**Behaviour** (obfuscated `\uXXXX` + reversed strings, runs on `postinstall: node test.js`):
- fetches an attacker SSH public key from C2 `http://170.205.31.203` and appends it to `~/.ssh/authorized_keys` (mode 0600) — persistent SSH backdoor
- `sudo ufw allow` (opens the firewall) + `sudo chown`
- `wmic logicaldisk get name` host recon
- exfiltrates discovered files to the same C2 (`/api/v1`)

**IOCs:** C2 `170.205.31.203` · tarball `sha256:97690f6e2d4ce315f19595b04cae3afdac0bfe9d98b0d3caeb1fd99eab133cde`

Not present in OSV or GHSA at time of reporting (first-catch). Behaviour confirmed by deterministic structural analysis of the published tarball. npm takedown filed in parallel.

PR #1365, opened 2026-07-09
Malicious packages: npm @wagni_bot campaign — 25 crypto typosquats (postinstall SSH-key + wallet + env stealer, Telegram exfil)

akyroslabs:
This PR reports a **coordinated campaign of 25 malicious npm packages** published under the single scope **`@wagni_bot`** on 2026-07-09, targeting crypto/web3 developers.

## Behaviour
Each package ships a `postinstall` hook (`node postinstall.js`) that runs on `npm install`, before import. It:
- fingerprints host/user (`os.hostname()`, `os.userInfo()`, `os.platform()`),
- walks the home directory and reads **SSH private keys (`~/.ssh/id_rsa`)**, **crypto wallet files**, and **`.env` secrets**,
- JSON-encodes the data and **exfiltrates it to a single hard-coded Telegram bot** (`sendMessage`).

## Why these belong together (deterministic)
- The payload file is **byte-identical across all 25 packages at each version** (version-lockstep; the actor re-published the same stealer across the scope, iterating 1.0.0 → 1.2.0).
- **All 25 share one Telegram exfil bot token** — a single automated actor.

## Packages (25)
metemask-sdk, metamask, binance-sdk, solana-sdk, jupiter-sdk, orca-sdk, meteora-sdk, pumpfun-sdk, opensea, opensea-sdk, polymarket, polymarket-sdk, hyperliquid, hyperliquid-sdk, polygon, polygon-sdk, bsc, bsc-sdk, eth, eth-agent, ethereum-wallet, web3, web3-agent, web3-toolkit, wagni — all under `@wagni_bot`.

Detected and classified independently from the live npm feed by **codelake Research** on 2026-07-09. At time of reporting, none of the 25 were present in OSV or GHSA (a first-catch). Per-package technical advisories: https://research.codelake.dev/advisories (CLR-2026-3000 … CLR-2026-3024). A takedown report has been filed with npm.

PR #1394, opened 2026-07-21
Add iocs.files[] to database_specific for file indicators (per #1345 discussion)

akyroslabs:
This implements the structured file-IoC shape @calebbrown proposed in the review of #1345, so that file indicators (hashes etc.) have a documented home — today `iocs` only defines `domains`/`ips`/`urls`, and file hashes have nowhere schema-defined to go.

`iocs.files[]` records each file as `{ path, note, source, digests:{sha256,tlsh} }`. The `source` enum (`package-archive` / `downloaded` / `generated`) keeps package artifacts separate from files dropped or generated at run time, as suggested. `path` is optional when a file is identified only by digest; each entry must carry a `path` or at least one digest.

Includes the `Indicators` struct + validation (source enum, SHA-256/TLSH format, length limits), unit tests, and `docs/schema_additions.md`.

Flagged as a proposal — happy to adjust the shape (e.g. digest algorithms, required fields) per contributor consensus, as noted in #1345.


docs/schema_additions.md
```diff
@@ -77,6 +83,54 @@ a scheme/authority should be present.
...
+
+- `package-archive` — the file is part of the published package artifact.
+- `downloaded` — the file was retrieved at run time (e.g. from a C2).
+- `generated` — the file was produced at run time from code in the archive or
+  data downloaded from a C2 (e.g. base64-decoded or decrypted).
+
+#### iocs.files[].digests field
+
+The `digests` field is an optional JSON object mapping a hash algorithm name to
+the file's digest under that algorithm. Recognised algorithms are:
```
kam193: Maybe let's extend it to hashes recognized by VirustTotal (MD5, SHA1, SHA256, TLSH, SSDEEP) and add that it's prefered to add at least SHA256?

docs/schema_additions.md
```diff
@@ -77,6 +83,54 @@ a scheme/authority should be present.
...
+limited to 512 characters.
+
+#### iocs.files[].source field
+
+The `source` field is an optional string recording where the file came from. It
+must be one of:
+
+- `package-archive` — the file is part of the published package artifact.
+- `downloaded` — the file was retrieved at run time (e.g. from a C2).
+- `generated` — the file was produced at run time from code in the archive or
```
kam193: Maybe `dropped` instead of the `generated`, and add `memory` as an option for memory-only artifacts? Though I'm not sure if we want to include memory artifacts under files

docs/schema_additions.md
```diff
@@ -77,6 +83,54 @@ a scheme/authority should be present.
...
+payload downloaded from a C2, or content decoded from data embedded in the
+archive). Keeping these apart avoids mixing, e.g., ten package tarballs and two
+dropped files into a single flat list.
+
+Each object may contain the following fields. An entry must have at least a
+`path` or one digest under `digests`.
+
+#### iocs.files[].path field
+
+The `path` field is an optional string giving the file name as observed, either
```
kam193: I think I'd suggest renaming it as `paths` and typing to the list - the same file can exist in multiple places.

docs/schema_additions.md
```diff
@@ -77,6 +83,54 @@ a scheme/authority should be present.
...
+- `generated` — the file was produced at run time from code in the archive or
+  data downloaded from a C2 (e.g. base64-decoded or decrypted).
+
+#### iocs.files[].digests field
+
+The `digests` field is an optional JSON object mapping a hash algorithm name to
+the file's digest under that algorithm. Recognised algorithms are:
+
+- `sha256` — a lowercase hex-encoded SHA-256 digest (64 hex characters).
+- `tlsh` — a hex-encoded [TLSH](https://tlsh.org/) fuzzy hash (70 hex
```
kam193: I'd suggest accepting any valid form of the hash (lower/upper/mixedcase for all, prefixed or not for TLSH) but performing normalization in the CI.

docs/schema_additions.md
kam193: I also wonder if it's better to keep the list closed or open. No hard opinion.

docs/schema_additions.md
```diff
@@ -77,6 +83,54 @@ a scheme/authority should be present.
...
+The `files` field is a JSON array of objects, each describing a single file
+associated with the malicious package.
+
+Files are recorded separately from the network indicators above, and each entry
+carries its own `source` so that consumers can distinguish package artifacts
+(the published tarball/wheel, which may only ever be processed as a stream) from
+files that are dropped or generated at run time (for example a second-stage
+payload downloaded from a C2, or content decoded from data embedded in the
+archive). Keeping these apart avoids mixing, e.g., ten package tarballs and two
+dropped files into a single flat list.
```
kam193: I think this paragraph could be omited.

docs/schema_additions.md
calebbrown: Agreed. `"md5", "sha1", "sha256", "tlsh", "ssdeep"` should all be acceptable. `"sha256"` is preferred and recommended at a minimum.

docs/schema_additions.md
calebbrown: This sounds reasonable.

docs/schema_additions.md
calebbrown: My preference is that they are all hex-encoded lowercase. Although I am okay if they are mixed or uppercase as that is an easy conversion.

The CI can transform them to lowercase hex as needed.

docs/schema_additions.md
calebbrown: The paragraph could definitely be improved, although I am still keen to keep package artifacts (i.e. the archives downloaded from registries) separate from other files associated with the malware.

docs/schema_additions.md
```diff
@@ -34,6 +34,12 @@ elaborated in the next section. All strings contain UTF-8 text.
 			"domains": [ string ],
 			"urls": [ string ],
 			"ips": [ string ],
+			"files": [ {
+				"path": string,
+				"note": string,
+				"source": string,
+				"digests": { "sha256": string, "tlsh": string }
```
calebbrown: The digests object should be reformatted to be spread across multiple lines for legibility.

docs/schema_additions.md
calebbrown: I'd prefer to not have unconstrained entries, or at least have a short conversation if a new value was desired.

As for `"dropped"` instead of `"generated"` how would that differ with `"downloaded"`, as these are technically dropped as well?

I'd also be okay with just having `"package-archive"`, `"dropped"` and perhaps `"in-memory"` as well to keep it simpler.

calebbrown (commented):
cc @awsactran @KunalSin9h 

It would be great to get your input here as well. Thanks!

docs/schema_additions.md
kam193: Yeah, my thoughts were also going in the similar direction, so I'd be for the simplified version with dropped and in-memory. Additional information could be hold in the note

docs/schema_additions.md
kam193: My idea was that this is explained later when describing the `source` field - or did I not understand something?

akyroslabs:
Thanks @kam193 and @calebbrown — pushed a commit addressing all points:

- `path` → `paths` (a list; the same file can appear in several places)
- simplified `source` to `package-archive` / `dropped` / `in-memory` (downloaded/generated collapse into `dropped`; `in-memory` for memory-only artifacts, extra detail in `note`)
- `digests` now accepts `md5`, `sha1`, `sha256`, `tlsh`, `ssdeep` (sha256 preferred and recommended at a minimum); hex digests are validated case-insensitively and normalized to lowercase, `ssdeep` is left as-is
- multi-lined the `digests` object in the overview block and tightened the `files[]` intro paragraph
- an entry now requires at least one `paths` entry or one digest

Happy to keep iterating if anything else needs adjusting.

akyroslabs:
One related thing while we finalize this: we also have several confirmed, novel malware reports open that all happen to carry a `database_specific.iocs` block — #1347, #1364, #1365, #1368, #1382, #1389, #1392, #1393 (all CI-green, DCO-signed, not in OSV/GHSA at detection).

If they're waiting on the iocs schema landing here, that's completely understandable — but since they're confirmed malware, I'd rather the reports themselves not be blocked on the schema discussion. Happy to do whatever unblocks them on your end:
- update the iocs to the agreed `iocs.files[]` shape once #1394 lands,
- adjust it now to whatever you prefer, or
- simply drop the `iocs` block so the plain reports can merge, and re-add indicators later.

Just let me know what's easiest — mainly want to make sure they don't fall through the cracks. Thanks!

KunalSin9h:
@calebbrown loos good, hashes and path are paramount here.

calebbrown:
> * update the iocs to the agreed `iocs.files[]` shape once [Add iocs.files[] to database_specific for file indicators (per #1345 discussion) #1394](https://github.com/ossf/malicious-packages/pull/1394) lands,
> * adjust it now to whatever you prefer, or
> * simply drop the `iocs` block so the plain reports can merge, and re-add indicators later.

As for your other reports I am happy to leave them as just "hashes" for now, with a view to replace them with files later.

calebbrown:
I'm generally happy with this change, but I need to review the changes in more detail. I will have more comments on Monday.

The code changes and overall spec are the most important to get right, we can always clean up the other wording later.

I would also like a quick review from @awsactran too.

awsactran:
Sorry for the late response as I am currently OOO. Overall I think this is a solid, well-scoped proposal. I do like that it stays inside database_specific rather than trying to alter OSV proper.

internal/report/indicators.go
```diff
@@ -59,10 +113,75 @@ func (i *Indicators) UnmarshalJSON(b []byte) error {
...
+	for _, p := range f.Paths {
+		if len(p) > maxPathLength {
+			return fmt.Errorf("%w file path too long (%d > %d)", ErrUnexpectedOSV, len(p), maxPathLength)
+		}
+	}
+	if len(f.Note) > maxNoteLength {
+		return fmt.Errorf("%w file note too long (%d > %d)", ErrUnexpectedOSV, len(f.Note), maxNoteLength)
+	}
+	d := f.Digests
+	hasDigest := d != nil && (d.MD5 != "" || d.SHA1 != "" || d.SHA256 != "" || d.TLSH != "" || d.SSDEEP != "")
```
calebbrown, line 142: I think the requirement should be for either MD5, SHA1 or SHA256 to be required. TLSH and SSDEEP are for content similarity checks, so they should not be one of the required.

internal/report/indicators.go
```diff
@@ -59,10 +113,75 @@ func (i *Indicators) UnmarshalJSON(b []byte) error {
...
+}
+
+// validateDigests validates each present digest and normalizes the hex ones to
+// lowercase in place. ssdeep is case-sensitive (base64-ish) and left as-is.
+func validateDigests(d *FileDigests) error {
+	if d == nil {
+		return nil
+	}
+	var err error
+	if d.MD5, err = normHex("md5", d.MD5, md5RE); err != nil {
```
calebbrown: It is worth noting that "normHex" will not be persisted at the moment. The indicators struct (inside DatabaseSpecific) is only currently used for validation.

That is "okay" at the moment, but the normalization can be done later. It may improve clarity to remove the mutation for now.

internal/report/indicators.go
```diff
@@ -59,10 +113,75 @@ func (i *Indicators) UnmarshalJSON(b []byte) error {
...
+		return err
+	}
+	if d.SSDEEP != "" && !ssdeepRE.MatchString(d.SSDEEP) {
+		return fmt.Errorf("%w invalid ssdeep digest '%s'", ErrUnexpectedOSV, d.SSDEEP)
+	}
+	return nil
+}
+
+// normHex validates a hex digest (case-insensitive) and returns it lowercased.
+func normHex(name, val string, re *regexp.Regexp) (string, error) {
```
calebbrown: This is function is doing two things. My preference is for the validation to be independent of the ToLower() as it makes it simpler to read and understand.

docs/schema_additions.md
```diff
@@ -77,6 +89,60 @@ a scheme/authority should be present.
...
+
+The `note` field is an optional free-text string describing the file. It is
+limited to 512 characters.
+
+#### iocs.files[].source field
+
+The `source` field is an optional string recording where the file came from. It
+must be one of:
+
+- `package-archive` — the file is part of the published package artifact.
```
calebbrown: This should probably be reworded as:

`- the file is extracted from the published package artifact`

This eliminates any confusion about it potentially including the artifact itself.

docs/schema_additions.md
```diff
@@ -77,6 +89,60 @@ a scheme/authority should be present.
...
 
+### iocs.files[] field
+
+The `files` field is a JSON array of objects, each describing a single file
+associated with the malicious package.
+
+Each entry carries a `source` so that package artifacts (the published
+tarball/wheel, which may only ever be processed as a stream) are kept separate
+from other files associated with the malware — a second stage dropped at run
+time, or a payload that only ever lived in memory.
```
calebbrown: I think this paragraph could be improved. It should make it clearer what it is for, and be obvious that the package artifacts themselves do not belong in this set.

calebbrown (changes requested):
Only some minor comments, but otherwise this looks good. Thankyou!

For reference I am intending within the next week or so to tighten the validation and make it possible to also mutate the DatabaseSpecific fields (this is to clean up CWEs, Indicators and other things).

Also, a structure for listing package artifacts themselves separately to IOCs will need proposed to support that usage.

docs/schema_additions.md
```diff
@@ -77,6 +89,60 @@ a scheme/authority should be present.
...
+
+#### iocs.files[].source field
+
+The `source` field is an optional string recording where the file came from. It
+must be one of:
+
+- `package-archive` — the file is part of the published package artifact.
+- `dropped` — the file was written to disk at run time (e.g. a second stage
+  retrieved from a C2, or content decoded from data embedded in the archive).
+- `in-memory` — the file only ever existed in memory and never hit disk.
```
calebbrown: Can these enums be
- "PACKAGE_ARCHIVE"
- "DROPPED"
- "IN_MEMORY"

This keeps the enums consistent with the OSV enums, and allows the move to protojson based parsing of DatabaseSpecific data as well.

akyroslabs:
Thanks @calebbrown — addressed the review:

- **Enums** renamed to OSV style: `PACKAGE_ARCHIVE` / `DROPPED` / `IN_MEMORY` (in `indicators.go`, the tests, and the docs), for consistency with the OSV enums and the move to protojson-based `database_specific` parsing.
- **Validation decoupled from normalization**: `normHex` is now `validateHex` (a pure check, no mutation); the lowercasing of the hex digests happens as a separate step in `validateDigests`.
- **Docs**: reworded `PACKAGE_ARCHIVE` to "the file is **extracted from** the published package artifact", and rewrote the intro paragraph so it's explicit that the published artifact itself does **not** belong in this set (it's already identified by `affected.package`; this field is for the individual files of interest).

Makes sense re: tightening validation, mutable `DatabaseSpecific`, and a separate structure for listing package artifacts — happy to help on those follow-ups.

calebbrown (approved):

PR #1399, opened 2026-07-23
Fix a report that included a special char.

calebbrown:
Unfortunately this means that the there is a package now with two reports. This change adds support for this situation.

Using the earliest file is not always the most desired case if there are multiple reports. In some cases it might be better to use the latest. But in this case this makes sense.

This change will also unblock ingesting into OSV.


kam193:
😱

distractible (approved):

PR #1405, opened 2026-07-28
report: npm/app-soda-layer 2.1.6 (credential + wallet stealer, SSH backdoor)

rbnfrhnr:
Adds a report for `app-soda-layer@2.1.6` (npm).

The package was published 2026-07-27T18:47:46Z and unpublished by its author 2026-07-27T21:13:07Z — an availability window of 2h25m. It is no longer retrievable from the registry, so the source in this report was reconstructed from a system-call trace of an installation performed in an isolated sandbox during that window.

**Behaviour**

- Runs on `npm install` via `postinstall`.
- Reads Solana CLI keypairs (`id.json`), `config.toml` and `.env` files and uploads their **contents**, with the HTTP client's size limits explicitly removed.
- Requests an SSH public key from its control server and appends it to `~/.ssh/authorized_keys`, corrects `~/.ssh` ownership so `sshd` will honour it under `StrictModes`, then attempts `ufw allow 22/tcp`. Implant success is reported back to the operator.
- The file patterns it collects are **not in the package** — they are fetched from the control server at install time, so the published artifact can be retargeted without republishing.

**Control server still active**

The three control endpoints on `95.216.118.146:3001` were requested directly and returned HTTP 200 on both 2026-07-28T06:45:57Z and 2026-07-28T11:57:24Z — roughly 9.5 and 15 hours after the package was withdrawn. Responses were byte-identical between the two fetches (sha256 recorded in the report). The infrastructure is operated, not abandoned.

Those requests were GET-only from a disposable host; no request body was sent, and the upload endpoints (`/api/v1`) were deliberately not contacted.

This means the report includes the **actual SSH public key** the malware implants, so anyone affected can test their own `authorized_keys` for it directly, and the **actual target list** (`seed`, `phrase`, `wallet`, `privatekey`, `.xls`/`.xlsx`, `.env`).

**Note on the key's comment field**

The retrieved key carries a comment naming a third-party company's support address. An SSH key comment is arbitrary, unvalidated text chosen by whoever generated the key; it is camouflage and is **not** evidence of involvement by that organisation. The report states this inline, immediately below the key, because the entry will syndicate and the disclaimer needs to travel with it.

**Conventions**

- `id` omitted (assigned automatically); `summary` omitted (overridden).
- `affected[].versions` pinned to `2.1.6` rather than an open range — the registry record shows the package name and its only version were created 284ms apart, so no other version ever existed.
- IOCs split into `ips` (host) and `urls` (full endpoints) per `docs/schema_additions.md`.

Full technical write-up, including the recovered source and the syscall evidence: https://protet.io/disclosures/app-soda-layer-2.1.6


rbnfrhnr:
Updated to redact the SSH key's comment field.

The retrieved key carried a comment impersonating the support address of a real, unrelated company. On reflection I don't think it belongs in a public malware record: naming the impersonated company attaches it to this entry for no detection benefit, since **the base64 body is the correct thing to match on regardless** — the comment can be changed without changing the key.

The key body itself is unchanged and complete. The report notes that the comment is redacted deliberately rather than unknown, and that the unredacted line stays retrievable from `/api/ssh-key` on the reported host for anyone who wants to verify it independently.

The corresponding write-up has been updated the same way.

osv/malicious/npm/app-soda-layer/MAL-0000-app-soda-layer.json
```diff
@@ -0,0 +1,60 @@
...
...ability window.\n\n## Concealment\n\nThe `readme.md` was copied verbatim from an unrelated, permissively licensed open-source project and bore no relation to the shipped code. The package declared no description, author, repository or homepage. `child_process` and `os` were listed as dependencies despite being Node.js built-ins. Windows enumeration suppressed subprocess output via `stdio: [\"ignore\", \"pipe\", \"ignore\"]`, `-NoProfile` and `windowsHide: true`. Every stage was wrapped in exception handlers that discard errors, so a failed collection would not cause a visibly failed install.\n\n## Remediation\n\nAny environment where this package appears in a lockfile, CI log or `node_modules` tree dated 2026-07-27 should be checked for the SSH key above in `~/.ssh/authorized_keys`, on every account — a match indicates host access rather than only credential exposure. Firewall state should be checked for port 22 having been newly permitted. Solana keypairs, SSH private keys and any secrets in `.env` files reachable from the install directory should be rotated. Because routine 2's targets were supplied by the control server, the file list above should not be treated as exhaustive.",
```
elitsa-gosst: Nit: can you please make the description more concise?

elitsa-gosst:
Hi, thanks for your contribution to the Malicious Packages repo!

We're always excited to have more contributors! :)

elitsa-gosst (changes requested):

rbnfrhnr:
Thanks for the review, and for the welcome!

The description indeed was far longer than anything else in the repo. I've cut it from ~6,200 to ~1,800 characters, keeping only what a consumer of the entry needs: what it does on install, the operator-supplied targeting (the file patterns aren't in the package, they're fetched from the C2 at install time), and the implanted SSH key as a checkable indicator.

Happy to trim further if it's still too long.

osv/malicious/npm/app-soda-layer/MAL-0000-app-soda-layer.json
rbnfrhnr: done

rbnfrhnr:
One more update, and this one changes the framing.

This payload is not new. It matches the `levex-refa` / `lint-builder` toolkit from the February 2026 dev-protocol GitHub organisation compromise, and the same toolkit was reported here against different infrastructure in #1366 (`polymarket-mcp-v2@2.1.6`, C2 `170.205.31.203`). The overlap is in the distinctive parts: the exported `from_str_1` / `from_str_2` names, the exact `authorized_keys` → `chown -R` → `ufw enable` → `ufw allow 22/tcp` sequence, the `/api/ssh-key` `/api/scan-patterns` `/api/block-patterns` `/api/v1` endpoint names on port 3001, routine 1's `id.json` / `config.toml` / `.env` list, and both July packages published as version `2.1.6`.

So the package and the C2 are new, the payload is a redeployment. I've added a short paragraph to `details` saying so, plus the StepSecurity article as a reference. Happy to trim elsewhere if that pushes it past what you'd like — it's now ~2.6k characters.

One difference worth recording: the sibling obfuscates identifiers with reversed string literals (`"nosj.di".split("").reverse().join("")`), whereas the source recovered here has none. Different builds of the same tool.

Also, the SSH public key in this report was not published in either prior write-up, so if it's useful to cross-reference against #1366 that comparison would establish a common operator directly rather than by code similarity.

elitsa-gosst:
Can you please trim it a bit more and make it more human-readable? Here's a recent example of a report, which is of a good quality:  #1395

We haven't yet published a guideline for contributing to the repo, especially in regards to using AI, but the reports should still be written with a human user in mind.

rbnfrhnr:
Trimmed the details to two prose paragraphs in the style of #1395 and dropped the inline link, since the writeup is already in references. This report complements the Amazon Inspector record already on the package rather than repeating it: it adds attribution to the levex-refa/lint-builder toolkit from the February 2026 dev-protocol org compromise (with the cross-reference to #1366, polymarket-mcp-v2), the injected SSH public key that the Amazon finding notes is fetched but does not publish, and the file-target list retrieved from the live C2. Happy to trim further if you would like.

elitsa-gosst (approved):

PR #1412, opened 2026-07-30
protojson injects random whitespace. Fix the problem.

calebbrown:
protojson deliberately injects random whitespace after the ":" character.

This change fixes the indenting so that it is consistent.

See https://github.com/golang/protobuf/issues/1121 for more details


calebbrown:
Currently new reports are fixed by the ID assigner. Merged reports are going to look a bit funky.

distractible (approved):

PR #1420, opened 2026-08-04
Add malicious package reports for 15 npm packages compromised on 2026-08-04

r-bedekar:
Adds reports for 15 npm packages compromised on 2026-08-04 by a self-propagating worm that ships an install-time dropper and a Bun-bundled second-stage payload. 71 versions across four maintainer organisations: `@servicetitan` (4 packages), `@or-sdk` and `@onereach` (6), `@ornikar` (2), and the Qlik picasso packages (3).

Campaign linkage is established by the payload rather than by attribution: the byte-identical 727680-byte artifact `sha256 9fc2570b7cef51c1b8df116d144d11ff4096357be7d2c4c6367cfc2509cf1bcc` appears in packages published the same day under several unrelated maintainer organisations. The same activity has been reported publicly by StepSecurity, which names it "ChainDrop", and by Socket. Neither writeup lists any of these 15 packages; both are linked from every report for correlation.

Verification performed for every version listed:

- the `preinstall` hook was read from the registry packument and matched `node setup.mjs` exactly;
- the immediately preceding release of each package was checked and carries no such hook, establishing that the files were introduced by these versions;
- the tarball for the latest affected version of each package was downloaded and its embedded payload hashed to the value above.

Two indicator notes are recorded in the reports:

- `sha256 fd3ca4007b225fdf8de7af4345a19179d5efa8c4bb9205f88cda806e5684b1eb` has been published elsewhere as a "repository loader variant". It was recovered here directly from the npm tarballs of these packages and is the npm install-time loader for this wave. The distinct loader `sha256 54dc7ea54a1317cca0e890a2770630cf7fa6c97813e0cb9d2caa93012b350668` appears in the keyv/cacheable wave of the same campaign. Both are npm loaders.
- The dropper filename is randomised per maintainer organisation, `math_init.js` in these packages and `Math_Symbol.js` in the keyv/cacheable wave, with identical payload bytes, so detections keyed to a single filename match only part of the campaign.

Reports for the keyv/cacheable and Qlik runtime packages already covered by other sources are intentionally omitted to avoid duplicate entries. At the time of filing all 15 packages were still serving a malicious `latest` tag.

### DIFF
--- a/osv/malicious/npm/@onereach/styles/MAL-2026-11656.json
+++ b/osv/malicious/npm/@onereach/styles/MAL-2026-11656.json
@@ -2,12 +2,12 @@
   "schema_version": "1.7.4",
   "id": "MAL-2026-11656",
   "published": "2026-08-04T13:22:10Z",
-  "modified": "2026-08-05T00:12:09.750607885Z",
+  "modified": "2026-08-05T04:45:20Z",
   "aliases": [
     "GHSA-8pw5-rhrc-pv3j"
   ],
   "summary": "Malicious code in @onereach/styles (npm)",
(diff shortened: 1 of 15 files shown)


calebbrown:
Thank you for your report, and I am sorry to have merged #1426 ahead of yours - however due to the urgency of having reports and the comprehensiveness of #1426 in terms of packages I chose to merge that PR first.

I would appreciate it if you were still able to add your credits and the SHA256 hashes to reports that are present.

r-bedekar:
Thanks for explaining, and no worries about the ordering. Given how fast this one was moving that was the right call.

I've reworked #1420 so it no longer duplicates anything. It now just adds the SHA-256 digests and my credit to the 15 reports that are already there. The existing records list the file paths but don't carry hashes, so that's the gap I've filled:

```
package/setup.mjs       11,017 bytes
  sha256  fd3ca4007b225fdf8de7af4345a19179d5efa8c4bb9205f88cda806e5684b1eb

package/math_init.js   727,680 bytes
  sha256  9fc2570b7cef51c1b8df116d144d11ff4096357be7d2c4c6367cfc2509cf1bcc
```

One small correction is in there too. The two dropper filenames aren't the same file renamed at runtime. Both ship inside the tarball, and which one you get depends on the maintainer org. This wave ships `math_init.js`, the keyv/cacheable packages ship `Math_Symbol.js`. The payload bytes are identical either way, so one hash covers the whole campaign but one filename doesn't.

Every version listed was checked the same way: preinstall hook read from the packument, predecessor confirmed clean, then the tarball pulled and the payload hashed.

`check-preprocess` is failing and I can't work out why. All the other checks pass and the annotation only says "exit code 2". I don't have Go set up locally so I can't run `preprocess-strict` myself. If it's rejecting these as unmergable, any pointer to what it expects would help and I'll reshape them.

r-bedekar:
Sorted this out, no action needed from you.

The error was `merge failure: reports both have user details, skipping ...`. I was adding new `MAL-0000-*` files alongside the reports that came in with #1426, and `preprocess-strict` won't merge two records that both carry user details. Obvious once I could actually see the message. The GitHub annotation only showed "exit code 2", so I installed Go and ran `make preprocess-strict` locally to get the real output.

The PR now edits the 15 existing `MAL-2026-*` reports directly instead of adding anything new. Each one gets my credit appended and the two SHA-256 digests added under `database_specific.iocs.files`. I left `details` and `affected` untouched since they're already accurate, and kept the filename note in my earlier comment rather than repeating it across 15 records.

All checks pass now.

calebbrown (approved):

calebbrown:
Thanks for the updates

PR #1437, opened 2026-08-11
Add 7 malicious packages: DeFi namesquat credential harvester + OAST /etc/shadow exfil

rohitcoder:
## Summary

Nine packages across two unrelated campaigns, both caught and sandbox-confirmed on 2026-08-11.

We sandbox-confirmed runtime execution and exfiltration behavior independently for all nine packages.

## Campaign 1 — DeFi namesquat credential harvester

Six npm and two PyPI packages published by the same actor within a ~27-minute window, all namesquatting real DeFi protocols and shipping a byte-identical install-time credential harvester.

### Affected packages

| Package | Version | Namesquat target |
|---|---|---|
| ethereum-vault-connector (npm) | 1.0.0 | Euler's Ethereum Vault Connector |
| @aerodrome-finance/contracts (npm) | 1.0.0 | Aerodrome Finance |
| @aerodrome-finance/slipstream (npm) | 1.0.0 | Aerodrome Finance Slipstream AMM |
| boring-vault (npm) | 1.0.0 | Seven Seas / Veda Boring Vault pattern |
| camelot-ammv2-core (npm) | 1.0.0 | Camelot DEX AMM v2 |
| camelot-ammv2-periphery (npm) | 1.0.0 | Camelot DEX AMM v2 |

### Attack chain

1. Actor registers a brand-new publisher account and ships packages under names/scopes impersonating well-known DeFi protocols with no prior legitimate npm/PyPI presence
2. Each npm package's `package.json` wires `preinstall`/`postinstall` to `node index.js || true`; each PyPI package runs equivalent logic via top-level `setup.py` exec and `__init__.py` import-time code — both trigger automatically on install, no explicit user action needed
3. The script scans `process.env` for variable names matching a broad credential regex (KEY, TOKEN, SECRET, PASS, PRIVATE, MNEMONIC, RPC, AWS, GITHUB, NPM, KUBE, VAULT, AUTH, PGP, GPG, SEED, WALLET)
4. It reads `~/.npmrc` and `~/.gitconfig`, and lists the contents of `~/.ssh`, `~/.foundry/keystores`, `~/.config/hardhat`, and `~/.config/gcloud`
5. Harvested data (hostname, username, cwd, platform, env vars, file contents, directory listings) is bundled into JSON and POSTed over HTTPS to a shared webhook.site endpoint, immediately followed by a raw TCP connect to the same host confirming delivery
6. The install script wraps everything in a bare try/catch so failures never surface to the user, keeping the exfiltration silent regardless of outcome

All eight packages share the exact same webhook.site callback UUID, differing only in the `pkg` field logged in the exfiltrated JSON body and the destination IP for the raw TCP callback.

### IOCs

- **C2 domain:** `webhook.site`
- **C2 path (shared across all 8 packages):** `/326b0891-2093-4800-a4c1-686ce3e07b09`
- **Payload file (npm):** `package/index.js` — SHA-256 digests are included per-package in each report's `database_specific.iocs.files`

### MITRE ATT&CK

T1195.002 — Supply Chain Compromise: Compromise Software Supply Chain
T1552.001 — Unsecured Credentials: Credentials In Files

### Affected package

| Package | Version |
|---|---|
| @telekom-ods/react-ui-kit (npm) | 2.6.9 |

### Attack chain

1. Package ships an install hook that spawns a shell
2. The shell reads `/etc/passwd` and `/etc/hosts`, runs `id`, and conditionally reads `/etc/shadow` if the installing process has read access
3. Collected output is sent as the `User-Agent` header of an HTTP POST to an out-of-band interaction (OAST) callback host
4. `whoami` and `hostname` output are embedded directly in the callback URL path, letting the operator fingerprint each compromised host and receive file contents in a single request without running a body-logging listener

Registry metadata showed the same publisher as prior releases under this scope (not a freshly registered account), suggesting a compromised maintainer account/token or a supply-chain injection into the release pipeline, rather than a simple namesquat. Unrelated to Campaign 1 — different technique, different infrastructure, different publisher history. The package has since been fully removed from npm.

### IOCs

- **C2 domain:** `d9t83osijf9n1gb62e4gxfioysijywqww.oast.me`
- **C2 URL pattern:** `http://d9t83osijf9n1gb62e4gxfioysijywqww.oast.me/telekom-ods/$(whoami)/$(hostname)/`

### MITRE ATT&CK

T1195.002 — Supply Chain Compromise: Compromise Software Supply Chain
T1552.001 — Unsecured Credentials: Credentials In Files
T1005 — Data from Local System

## Reporter

O3 Security — https://o3.security


osv/malicious/pypi/dlmm-sdk/MAL-0000-dlmm-sdk.json
```diff
@@ -0,0 +1,54 @@
...
+      ]
+    }
+  ],
+  "database_specific": {
+    "iocs": {
+      "domains": [
+        "webhook.site"
+      ],
+      "ips": [
+        "178.63.67.153"
```
kam193: This is the official webhook.site IP, it is not dedicated to malicious activity. I don't think it's helpful to share it as IoC.

osv/malicious/pypi/dlmm-sdk/MAL-0000-dlmm-sdk.json
```diff
@@ -0,0 +1,54 @@
...
+      "type": "FINDER",
+      "contact": [
+        "https://o3.security"
+      ]
+    }
+  ],
+  "database_specific": {
+    "iocs": {
+      "domains": [
+        "webhook.site"
```
kam193: webhook.site itself is a legitimate site, I'd suggest not to include it as a generic IoC.

osv/malicious/pypi/dlmm-sdk/MAL-0000-dlmm-sdk.json
```diff
@@ -0,0 +1,54 @@
...
...ion confirmed the package reads `.npmrc` from the home directory (consistent with the same credential-harvesting template used by this actor's npm-side packages, which also scrape environment variables and SSH/Foundry/Hardhat/gcloud credential paths) and performs an outbound HTTP GET to a webhook.site endpoint immediately followed by a raw TCP connection to the same host on port 443 — an out-of-band callback confirming successful exfiltration to the operator.\n\nThis package shares the exact same webhook.site callback path as five other packages from the same actor published within the same short window on 2026-08-11 (`ethereum-vault-connector`, `@aerodrome-finance/contracts`, `@aerodrome-finance/slipstream`, `boring-vault`, `camelot-ammv2-core`, `camelot-ammv2-periphery` — all npm, all namesquatting real DeFi protocols) and a sibling PyPI package `dlmm`, also namesquatting Meteora's DLMM SDK. The package has since been removed from PyPI.\n\nAny environment where this package was installed should be treated as having had its environment and reachable credential files exposed to the attacker; rotate all reachable secrets, especially PyPI/npm publish tokens and any wallet/RPC keys.",
```
kam193: > followed by a raw TCP connection to the same host on port 443 

Do you mind reviewing this statement? This package is pretty simple; it has one place generating the request (`__init__.py`) invoked twice: in `setup.py` and when importing the module. I'd suggest your sandbox did not catch the second request properly.

kam193: > and performs an outbound HTTP GET to a webhook.site endpoint

I've just checked again, and in my environment, all requests were POST. Do you mind verifying your logs? It seems a bit like your analysis misreads POST for GET and then sees the data sent in POST as new "raw TCP" connection

osv/malicious/pypi/dlmm/MAL-0000-dlmm.json
```diff
@@ -0,0 +1,54 @@
...
...confirmed the package reads `.npmrc` from the home directory (consistent with the same credential-harvesting template used by this actor's npm-side packages, which also scrape environment variables and SSH/Foundry/Hardhat/gcloud credential paths) and performs an outbound HTTP GET to a webhook.site endpoint immediately followed by a raw TCP connection to the same host on port 443 — an out-of-band callback confirming successful exfiltration to the operator.\n\nThis package shares the exact same webhook.site callback path as five other packages from the same actor published within the same short window on 2026-08-11 (`ethereum-vault-connector`, `@aerodrome-finance/contracts`, `@aerodrome-finance/slipstream`, `boring-vault`, `camelot-ammv2-core`, `camelot-ammv2-periphery` — all npm, all namesquatting real DeFi protocols) and a sibling PyPI package `dlmm-sdk`, also namesquatting Meteora's DLMM SDK. The package has since been removed from PyPI.\n\nAny environment where this package was installed should be treated as having had its environment and reachable credential files exposed to the attacker; rotate all reachable secrets, especially PyPI/npm publish tokens and any wallet/RPC keys.",
```
kam193: See the same comments as in dlmm-sdk

osv/malicious/pypi/dlmm/MAL-0000-dlmm.json
```diff
@@ -0,0 +1,54 @@
...
+    }
+  ],
+  "database_specific": {
+    "iocs": {
+      "domains": [
+        "webhook.site"
+      ],
+      "ips": [
+        "178.63.67.106"
+      ],
```
kam193: See the same comments as in dlmm-sdk

osv/malicious/pypi/dlmm/MAL-0000-dlmm.json
```diff
@@ -0,0 +1,54 @@
...
+            "name": "Embedded Malicious Code"
+          }
+        ]
+      }
+    }
+  ],
+  "references": [
+    {
+      "type": "PACKAGE",
+      "url": "https://pypi.org/project/dlmm/"
```
kam193: I personally don't see any point in referencing to PyPI:
1) OSV.dev will add a link anyway based on the purl
2) you said the package was already removed, so the link is expected to be dead :)

osv/malicious/pypi/dlmm-sdk/MAL-0000-dlmm-sdk.json
rohitcoder: Fair point, that's the shared webhook.site IP, not attacker infra. Dropped it, kept the unique callback path since that's the actual indicator.

osv/malicious/pypi/dlmm-sdk/MAL-0000-dlmm-sdk.json
rohitcoder: Agreed, removed the bare domain from iocs across all 8 packages in this campaign, not just this one. Left the specific callback URL in place.

osv/malicious/pypi/dlmm-sdk/MAL-0000-dlmm-sdk.json
rohitcoder: You're right, checked our sandbox capture again and it's a single POST — the sandbox was logging the socket connect and the HTTP request as two separate events and I narrated them as sequential actions. Rewrote the details to say one POST, invoked from both setup.py and __init__.py.

rohitcoder: Confirmed, it's POST not GET (matches what we saw on the npm side of this campaign too). Fixed in both dlmm and dlmm-sdk. Thanks for catching that.

osv/malicious/pypi/dlmm/MAL-0000-dlmm.json
rohitcoder: Same fix applied here — dropped the bare webhook.site domain/IP from iocs and corrected the details to a single POST.

osv/malicious/pypi/dlmm/MAL-0000-dlmm.json
rohitcoder: Same as the dlmm-sdk thread — fixed.

osv/malicious/pypi/dlmm/MAL-0000-dlmm.json
rohitcoder: Good point on both counts, removed the PyPI PACKAGE link from dlmm and dlmm-sdk.

osv/malicious/pypi/dlmm/MAL-0000-dlmm.json
rohitcoder: i checked logs, events make sense but post/get needs to be checked

image

rohitcoder:
Hi @kam193 

Turns out dlmm and dlmm-sdk were already reported by by you as [MAL-2026-13728](https://github.com/ossf/malicious-packages/blob/main/osv/malicious/pypi/dlmm/MAL-2026-13728.json) and [MAL-2026-13729](https://github.com/ossf/malicious-packages/blob/main/osv/malicious/pypi/dlmm-sdk/MAL-2026-13729.json) before this PR landed - dropped both from here to avoid the duplicate, and updated the remaining npm reports to point at those IDs instead of re-describing them. Down to 7 packages now (6 npm + the telekom-ods one).

osv/malicious/pypi/dlmm/MAL-0000-dlmm.json
rohitcoder: @kam193 I removed these all pypi pacakges finding, no pointing in creating dupes

kam193:
You can upload yours as well - the CI will merge reports

calebbrown (approved):

PR #1450, opened 2026-08-16
Add OSV reports: AI-agent CLI C2 npm campaign (3 packages)

johncassil:
This adds three OSV malicious-package entries for a coordinated npm campaign, reported by pkgwarden after byte-level verification of the published artifacts.

**Entries**

- `oh-aicoding-tool` (0.1.74, 0.1.76, 0.1.77): installs a persistent Claude Code `Stop` hook (via scripts/langfuse-setup.mjs, reachable from bin/cli.js) that copies langfuse_hook.py into ~/.claude/hooks/ and registers it in ~/.claude/settings.json, with a parallel Codex/OpenCode setup path. On every agent turn it reads the AI-session transcript and git repository context and exfiltrates them using hardcoded shared attacker Langfuse keys.
- `oh-aireport` (0.2.2): bundles an OpenCode plugin (.opencode/plugins/oh-ai-report.ts) whose `DEFAULT_WEBHOOK_URL` is the attacker C2, shipping session and git context plus user_id and company user_email by default with no configuration.
- `upshift-sdk` (8.20.1): package.json `scripts.postinstall` unconditionally runs an inline `node -e` https.get to an attacker Cloudflare Worker on every install, exfiltrating hostname, username, cwd, and package name/version, silenced with try/catch. Same C2 host and exact 8.20.1 version string as the already-confirmed-malicious augustdigital-sdk@8.20.1 (MAL-2026-13774), upshift-config (MAL-2026-13776), and upshift-finance (MAL-2026-13777).

**IOC hygiene**: the oh-* C2 is the attacker-operated raw IP 120.46.221.227 (fronted by metrics.openharmonyinsight.cn under the "openharmonyinsight" branding); full C2 paths are in `urls`. The upshift-sdk worker host is `*.workers.dev` shared hosting, so it is recorded only as a full URL in `urls`, never as a bare domain.

Reported by pkgwarden (https://pkgwarden.com).

---

**Related analysis:** these three packages surfaced from pkgwarden's cross-family C2 clustering of a broader coordinated campaign (~75 OSV MAL families of trojanized/impersonated AI-coding-agent CLIs). Full writeup, with the shared-infrastructure clusters and how these three were confirmed: https://pkgwarden.com/incidents/ai-agent-cli-c2-npm-august-2026/

### DIFF
--- a/osv/malicious/npm/upshift-sdk/MAL-0000-upshift-sdk.json
+++ b/osv/malicious/npm/upshift-sdk/MAL-0000-upshift-sdk.json
@@ -1,0 +1,52 @@
+{
+  "schema_version": "1.7.4",
+  "modified": "2026-08-16T00:00:00Z",
+  "published": "2026-08-16T00:00:00Z",
+  "summary": "Malicious code in upshift-sdk (npm)",
+  "details": "upshift-sdk exfiltrates host reconnaissance to attacker infrastructure from a postinstall script that runs on every npm install. package.json declares scripts.postinstall as an inline `node -e` program that unconditionally performs an https.get to an attacker-controlled Cloudflare Worker, sending os.hostname(), os.userInfo().username, the current working directory, the package name and version, and a timestamp. The request is wrapped in try/catch so failures stay silent and leave no obvious error. The collector is https://build-metrics-collector.cdn-ops-health.workers.dev/npm-install/. This package is part of the same campaign as the already-confirmed-malicious upshift-config, upshift-finance, and augustdigital-sdk: it uses the same command-and-control host and carries the exact same 8.20.1 version string as augustdigital-sdk@8.20.1, indicating a single actor publishing coordinated postinstall-exfiltration packages. See the related advisories referenced below.",
+  "affected": [
+    {
+      "package": {
+        "ecosystem": "npm",
+        "name": "upshift-sdk"
+      },
+      "versions": [
+        "8.20.1"
+      ]
+    }
+  ],
+  "references": [
+    {
+      "type": "PACKAGE",
+      "url": "https://www.npmjs.com/package/upshift-sdk"
+    },
+    {
+      "type": "ADVISORY",
+      "url": "https://osv.dev/vulnerability/MAL-2026-13774"
+    },
+    {
+      "type": "ADVISORY",
+      "url": "https://osv.dev/vulnerability/MAL-2026-13776"
+    },
+    {
+      "type": "ADVISORY",
+      "url": "https://osv.dev/vulnerability/MAL-2026-13777"
+    }
+  ],
+  "credits": [
+    {
+      "name": "pkgwarden",
+      "type": "FINDER",
+      "contact": [
+        "https://pkgwarden.com"
+      ]
+    }
+  ],
+  "database_specific": {
+    "iocs": {
+      "urls": [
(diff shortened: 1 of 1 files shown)


osv/malicious/npm/oh-aicoding-tool/MAL-0000-oh-aicoding-tool.json
calebbrown, line 1: After reviewing the payload described in this detail field I am trying to understand why `oh-aicoding-tool` and `oh-aireport` are considered malware, over a poor attempt at implementation telemetry collection.

1. The affected versions appear to be more than those listed below. All versions I inspected (from 0.1.0 to 0.2.6) have the same IP address, domains and hardcoded keys in the listed script.
2. The code `const publicKey = args.publicKey || process.env.LANGFUSE_PUBLIC_KEY || "pk-lf-da0c9...";` suggests that this is a default value, as the user's one keys would take precedence.
3. IP address 120.46.221.227 appears to be used throughout the codebase without any attempt to mask it's purpose or behavior.
4. The persistent hook and telemetry is documented in the README.

Exfiltrating AI session context, repo details, and current working directory are definitely problematic though.

johncassil: Thanks for the close read; you're right on all four points, and we dug further. The IP is Huawei Public Cloud (HWCSNET), the README gates on a Huawei-format employee ID (`^[a-z](?:\d{8}|wx\d{7})$`, e.g. `h00613222`/`hwx1234567`), and the whole suite (including `oh-langfuse`, which our shared-infrastructure pivot anchored on) appears to be an OpenHarmony-program internal telemetry toolkit published to public npm. There is no postinstall; setup is an explicit command or an opt-in interactive menu (items unchecked by default). The `oh-aireport` webhook fires only on an explicit `/report-ai-issue` invocation, so our "by default" wording overstated it. We also confirm the affected-version issue: every version from 0.1.0 through 0.2.7 (published this morning, about an hour after your review, with the same IP and keys) carries the same infrastructure.

What remains true: once configured, the hook ships full user prompts, assistant responses, and tool inputs/outputs (which routinely include file contents and any secrets observed in tool calls) over cleartext HTTP to a bare IP, pre-authenticated with a shared secret key baked into a public package, which appears to let anyone holding the key read every user's collected traces. Amazon Inspector's MAL-2026-4625 classifies this same behavior in `oh-langfuse` as malicious today, which is the precedent our entries followed.

We'll defer to you on classification. Happy to either (a) drop the two `oh-*` entries from this PR and keep `upshift-sdk`, or (b) keep them with version ranges expanded to all versions if you consider the `oh-langfuse` precedent controlling. If (a), MAL-2026-4625 may deserve the same reassessment.

calebbrown: @awsactran for `oh-langfuse` - did AWS identify anything specifically malicious?

johncassil:
Following up on our own reassessment above: we have withdrawn the two oh-* entries (oh-aicoding-tool, oh-aireport) from this PR rather than wait on the oh-langfuse question. The branch is now a single signed commit adding only upshift-sdk, which stands on its own evidence. Apologies for leaving the withdrawn entries up for two weeks.

PR #1458, opened 2026-08-20
Add OSV reports for arrayref build-time payload campaign - crates.io (9 packages)

KunalSin9h:
Reports for a coordinated crates.io campaign observed on 2026-08-20.

**Trojanized releases of legitimate crates** (published from a compromised maintainer account): `arrayref` 0.3.10, `internment` 0.8.7, `append-only-vec` 0.1.9. Each adds a dependency on an attacker-controlled crate whose build script downloads and runs a remote binary at build time. Earlier releases are unaffected.

**Malicious dependency crates that carried the payload** (all versions): `proc-macro1` (a `proc-macro2` typosquat that fetches and executes a binary from `23.254.165.112` at build time), plus `proc-macro-en`, `aovine`, `arone`, `aronenao`, `tinymember`.

More detail: https://safedep.io/arrayref-proc-macro1-rust-build-time-malware/


osv/malicious/crates.io/proc_macro1/MAL-0000-proc_macro1.json
```diff
@@ -0,0 +1,113 @@
...
+        },
+        {
+          "source": "PACKAGE_ARCHIVE",
+          "paths": [
+            "proc-macro1-1.0.106.crate"
+          ],
+          "digests": {
+            "sha256": "b5c1b5b0763a8809a644a8f92224653f0aca623a98eecc714d27f74b80fbe436"
+          }
+        },
```
calebbrown, line 88: The "source" is for files inside the package archive, not the archives themselves.

I intend to publish another data structure for capturing these soon.

calebbrown (changes requested):
Some of the packages reported like `aovine` and `arone` don't appear to be malicious. Are they merely packages that were uploaded by the same account that uploaded proc_macro1?

calebbrown (approved):
I'm going to merge this anyway so there are reports.

PR #1466, opened 2026-08-24
Move database specific parsing to a protobuf

calebbrown:
Add a proto for parsing database specific.

- Vuln level Origins and IOCs are validated.
- Serialising is frustrating - origins are manually pushed back into the structpb.Struct.
- Report.Validate now exists to simplify tests.
- `make gen-protos` and GitHub Actions now exist to make it easier to manage.

Not done yet:
- full validation across the entire proto
- normalization persisted of iocs
- proto linting


Makefile
```diff
@@ -32,6 +32,10 @@ test/unit:
 validate: ## Validate all OSV files
 	go run ./cmd/validate -config ./config/config.yaml
 
+.PHONY: gen-protos
```
distractible: this should be `generate` I guess

distractible (commented):
Are there unfinished (from the original message) that still need to be done? Is there an issue to link as well?

Makefile
calebbrown: True. Fixed.

Issue #1448, opened 2026-08-14
Correct the affected version range for MAL-2026-1383


chumil69:
## Request

Please update `MAL-2026-1383` to scope the finding to the specific version and test artifact that were identified, rather than the entire `@immuta/pxl-components` package.

## Background

The affected package is:

- Package: `@immuta/pxl-components`
- Specific version: `99.99.0`
- OSV record: [MAL-2026-1383](https://osv.dev/vulnerability/MAL-2026-1383)
- SafeDep report: [01KKJTXE78BNTXQGTPY149GRCM](https://app.safedep.io/community/malysis/01KKJTXE78BNTXQGTPY149GRCM)

Version `99.99.0` was published temporarily as part of an authorized, Immuta-initiated Bugcrowd penetration test designed to demonstrate a malicious-code injection scenario. It was a security-testing artifact, not a production release or a package shipped to Immuta customers. The package is no longer publicly available on npm.

The current OSV record uses an affected range beginning with:

```json
"introduced": "0"

elitsa-gosst:
This is now done.

PR #1467, opened 2026-08-24
Correct affected ranges

elitsa-gosst:
Update MAL-2026-1383 to specify the 2 malicious versions according to the request in #1448 . The package is no longer public on NPM.


kam193:
I think you have updated the wrong entry (historical key from the original report instead of the main key used by OSV)

elitsa-gosst:
> I think you have updated the wrong entry (historical key from the original report instead of the main key used by OSV)

You're absolutely correct! And I thought I double-checked that. :)

calebbrown (approved):

PR #1481, opened 2026-08-31
Add MAL-0000-13311.json for malicious fuels-forc package

loganaden:


osv/malicious/npm/fuels-forc/MAL-0000-13311.json
```diff
@@ -0,0 +1,28 @@
+{
+  "schema_version": "1.6.0",
+  "summary": "Malicious code in fuels-forc",
+  "details": "The package 'fuels-forc' was identified as containing malicious obfuscated code.",
```
kam193, line 4: A personal note: While there are reports with generic descriptions (and I also use them), maybe you would like to add a bit more, e.g. 1-2 sentences, describing what the payload does (if you know that)? A report with a brief description of the behavior is much more useful in further analysis, and I think it's good to write something more when the report is manual.

calebbrown: Yes please. For future reports this would be great!

calebbrown (approved):

PR #1436, opened 2026-08-10
Add OSV reports: DGXeon WhatsApp newsletter auto-follow (4 npm packages)

johncassil:
## Summary

Adds four novel npm malware reports from the DGXeon / dreamguyxeon campaign (consentless remote WhatsApp newsletter FOLLOW via GitHub C2). Distinct from existing OSV entries `dgxeon-baileys` (MAL-2026-2252) and `baileys-dgxeon` (MAL-2025-806).

| Package | Versions | Behavior |
|---|---|---|
| `my-auto-follow` | 1.0.0–1.0.3 | Runtime H29 in `lib/Socket/newsletter.js`: 120s → `DGXeon13/strings` → silent FOLLOW |
| `@dreamguyxeon/baileyx` | 1.0.0–5.0.0 | Same; also aliases `libsignal` → `@dgxeon13/libsignal-node`. LPM Firewall already public-malicious on 2.0.0 |
| `@dgxeon13/libsignal-node` | 1.0.0 | On `require`, overwrites `@whiskeysockets/baileys` `newsletter.js` with the same payload |
| `@dreamguyxeon/libsignal-node` | 1.0.1 | Same import-time Baileys patcher (aliased from `my-auto-follow`) |

**Claim discipline:** consentless remote-controlled newsletter FOLLOW / supply-chain tampering of Baileys. Not claiming credential theft or pairing-code device takeover from these tarballs.

## Evidence

- Live C2 (verified 2026-08-10): `https://raw.githubusercontent.com/DGXeon13/strings/refs/heads/main/strings.json` → `["120363409985606068@newsletter"]`
- Corroboration: [LPM Firewall `@dreamguyxeon/baileyx@2.0.0`](https://firewall.lpm.dev/npm/@dreamguyxeon/baileyx/v/2.0.0)
- Corroboration: [SafeDep — malicious Baileys npm WhatsApp campaign](https://safedep.io/malicious-baileys-npm-whatsapp-campaign/)
- Related: [MAL-2026-2252](https://osv.dev/vulnerability/MAL-2026-2252), [MAL-2025-806](https://osv.dev/vulnerability/MAL-2025-806)
- npm malware reports also filed for these four packages
- Finder credit: **pkgwarden**

## Test plan

- [ ] Validation GitHub Action passes
- [ ] Confirm no `id` / `summary` fields (auto-assigned)
- [ ] One affected package per file; versions match live npm

PR #1483, opened 2026-09-02
Add OSV reports: baileys newsletter auto-follow forks, systemzero + mrlegendbot (2 packages)

johncassil:
Two more npm forks of @whiskeysockets/baileys carrying the WhatsApp newsletter auto-follow behaviour already in the feed as MAL-2026-13929 and MAL-2026-13932 (PR #1436; SafeDep write-up: https://safedep.io/malicious-baileys-npm-whatsapp-campaign/).

To answer the question from #1436 up front: both packages take actions on the victim's WhatsApp account that are chosen remotely by a third party, are not disclosed anywhere in the package, and cannot be turned off by the developer using the library.

- `@systemzero/baileys@1.1.2`: `makeWASocket()` unconditionally opens a server-sent-events stream to a hardcoded host (`zone.api.br`) and executes remote `follow` / `react` commands on the victim's authenticated session (`lib/Socket/index.js` line 4 and 14, `lib/Utils/signal-stream.js` lines 48 to 69). A live command channel rather than a static list, so stronger than the merged entries.
- `@mrlegendbot/baileys@1.2.4`, `1.2.5`: `startAutoFeatures()` is scheduled at socket construction, fetches a channel list from a GitHub raw URL and follows every id on a 10 second loop (`lib/Socket/newsletter.js` lines 230 to 255, byte-identical across both versions).

Both diffed against upstream 7.0.0-rc14; the additions have no upstream equivalent. Two independent static reviews each; nothing was executed and the control URLs were not requested. Per the #1436 review, the GitHub raw URL is listed under `urls` only and no `domains` entry is set.


calebbrown (approved):

calebbrown:
Thanks for the report. I believe the versions can be expanded based on a brief inspection.

PR #1496, opened 2026-09-04
Add malicious package report for tsshare

loganaden:
base64 encoded url in client.py: "aHR0cHM6Ly9mc3p6dzU2LmNvbQ==".

IOC: https://fszzw56[.]com


calebbrown:
Can you provide a few more details on why you think this is malicious?

This looks like a borderline malicious case to me.

calebbrown (approved):

PR #1497, opened 2026-09-05
Fix description to reflect current state

kam193:
The advisory about this package was originally published with a wrong campaign assigned. The current merging strategy kept the old description after the update, requireing manual fix.

@calebbrown Is it the correct way to fix the description chosen by the merging strategy?


calebbrown (approved):

calebbrown:
This is fine for fixing details.

PR #1503, opened 2026-09-07
Add malicious npm package cat-sis2go-utils (dependency confusion, install-time beacon)

OGAREE:
Adds a report for `cat-sis2go-utils` (npm), versions 99.0.0 and 99.1.0. Both versions are still live on the registry.

The package ships no library code — `index.js` is an 87-byte stub whose own note reads "cat-sis2go-utils dependency-confusion PoC shim". The only executable content is the install payload: `package.json` declares both `preinstall` and `postinstall` running `node scripts/run.js`, which resolves a DNS canary under `dnshook.site` and POSTs to a `webhook.site` collector. 99.0.0 sends only a static marker; 99.1.0, published four minutes later, adds `os.userInfo().username` and `os.hostname()`.

The name combines `cat-` with `sis2go`, matching Caterpillar's SIS2GO product, and the versions are inflated to 99.x. Whether a private package under this name exists could not be verified from public data, so the report states that limitation rather than asserting it.

The package self-identifies as a proof of concept and the author field carries a wearehackerone.com address. It is submitted under the "security researcher activity" and "dependency and manifest confusion" scope items, on the same basis as the existing `box-sign-client` report, since the beacon fires on any install that resolves the package.

Report includes per-file SHA-256 digests, tarball SHA-1/SHA-256/SHA-512-SRI, and the observed IoCs. `id` and `summary` are omitted per CONTRIBUTING.

### DIFF
--- a/osv/malicious/npm/cat-sis2go-utils/MAL-0000-cat-sis2go-utils.json
+++ b/osv/malicious/npm/cat-sis2go-utils/MAL-0000-cat-sis2go-utils.json
@@ -1,0 +1,99 @@
+{
+  "schema_version": "1.7.4",
+  "modified": "2026-09-07T15:51:49Z",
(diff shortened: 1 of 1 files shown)


OGAREE:
Closing to resubmit from a clean branch. Will reopen shortly.

PR #1500, opened 2026-09-07
Add malicious PyPI package tsshare (Tushare impersonation; covert token + fingerprint exfil)

akyroslabs:
Adds the malicious PyPI package `tsshare`. Not previously in OSV or GHSA — **codelake first-catch** (the same package is independently flagged in #1496).

`tsshare` impersonates the popular Chinese market-data library **Tushare** — its docstrings advertise compatibility with `Tushare pro_bar` and it ships `MyShareClient`/`MyShareError` aliases (confusable name). Its default API endpoint is **deliberately concealed**: `tsshare/client.py` base64-decodes `aHR0cHM6Ly80Ny4xMTIuMTkxLjc1` to the hardcoded raw IP `https://47.112.191.75` (Alibaba Cloud, China). Legitimate clients do not base64-encode their own base URL.

On API use the client builds a persistent cross-platform **hardware fingerprint** (Windows `Win32_DiskDrive` serial via PowerShell, macOS `Hardware UUID` via `system_profiler`, Linux `/etc/machine-id`) and routes the request — carrying the user's Tushare `auth_code` token — to that hidden endpoint. There is **no install-time hook**; exfiltration occurs on API use, so it evades install-time scanners.

All eight versions in our corpus (1.0.5 → 1.0.19) carry the identical hidden endpoint and fingerprinting from the earliest version, so the entire package is malicious (`introduced: 0`).

**Advisory:** https://research.codelake.dev/advisories/clr-2026-3053-tsshare

Analysis method: static code + dataflow review of the published PyPI sdists; no execution, no live callback.

PR #1506, opened 2026-09-08
Update MAL-2026-16044 (tsshare): codelake details, IOCs, references, credit (parallel discovery)

akyroslabs:
Per @calebbrown's review on #1500 — updating the existing merged record `MAL-2026-16044` (originally reported by Cyberstorm.mu) rather than adding a duplicate. This was an independent parallel discovery by codelake Research.

Adds to the record:
- **Details** — full technical write-up: `tsshare` impersonates the `tushare` market-data library and conceals its default endpoint by base64-decoding it at runtime. In **1.0.5–1.0.18** it decodes to the hardcoded raw IP `https://47.112.191.75` (Alibaba Cloud, China); in **1.0.19** it rotates to the domain `https://fszzw56.com`. On use it builds a cross-platform hardware fingerprint (Windows `Win32_DiskDrive` serial, macOS `Hardware UUID`, Linux `/etc/machine-id`) and routes the user's Tushare `auth_code` token to that hidden endpoint. No install-time hook — exfiltration is on API use.
- **IOCs** — `ips` (`47.112.191.75`), `domains` (`fszzw56.com`), and `files[]` with the per-version `client.py` sha256 + the base64 endpoint literal in each note.
- **References** — codelake advisory.
- **Credits** — adds codelake Research as a second FINDER (parallel discovery).
- **Affected** — the hidden endpoint + fingerprinting are present from the earliest version we hold (1.0.5), so the whole package is marked malicious (`introduced: 0`).

Thanks @calebbrown for merging the first report and flagging this — supersedes #1500.


akyroslabs:
@calebbrown — first, a genuine apology. You were right on every point, and I appreciate you taking the time to spell it out rather than just bouncing it.

- **This PR (#1506)** updates the existing `MAL-2026-16044` instead of duplicating it: full technical `details`, IOCs — including the base64-encoded endpoint, which (good catch) **rotates from the raw IP `47.112.191.75` to the domain `fszzw56.com` in 1.0.19** — a reference to our advisory, and codelake added as a second FINDER credit alongside the Cyberstorm.mu team. The "not present in OSV or GHSA at report time" wording is gone.
- I've also gone back and **removed the same attribution / first-catch boilerplate from the `details` of all our other previously-merged records** — that's **#1507** (16 records, `details` prose only; no IOCs, versions, or credits changed).
- Most importantly, the root cause is fixed: our tooling no longer writes provenance/attribution into `details` at all — that belongs in `credits` / `references`, and the website carries any "first-catch" framing. This won't recur.

Thanks for merging the first report and for the patience on this one. Sorry for the extra review load.

calebbrown (approved):

PR #1507, opened 2026-09-08
Clean up details: remove codelake attribution boilerplate from 16 records

akyroslabs:
Follow-up to @calebbrown's review on #1500 / #1506.

The OSV `details` field should carry only the technical malware description, not provenance/attribution. This cleans up our previously-merged records by removing the trailing **"Detected and classified independently by codelake Research (a first-catch; not present in OSV or GHSA at report time)"** boilerplate from the `details` prose — that framing belongs in `credits` (which already carry the finder), not in the record body, and the "not present in OSV" claim was in some cases inaccurate.

16 records updated (details prose only — no technical content, IOCs, versions, or credits changed):

`@across-toolkit/typescript-config`, `@types-beta/sdk`, `@yancyyu/agentcli`, `agent-free`, `cc-skills-helper`, `codebuff-cli`, `date-format-utils-xz`, `datefmt-helper`, `faust-cont`, `fmt-date-lite`, `logger-daemon-regex`, `orbitron-cli`, `orbitron-tui`, `prime-coding-agent`, `supplyhub`, `tailwind-gutenberg-block-zero`.

For the two version-extension records (`@yancyyu/agentcli`, `codebuff-cli`) the self-attribution was reworded out while keeping the version-scope facts (which versions this report adds relative to the existing Amazon Inspector record).

Thanks again @calebbrown — this won't recur; our tooling no longer emits attribution into `details`.


calebbrown (changes requested):
I appreciate the clean up - this is super helpful.

However, can you please preserve the details after the `\n---\n_-= Per source details. Do not edit below this line.=-_\n` token in the string.

akyroslabs:
Good catch, thanks — fixed and force-pushed.

I re-did the cleanup from the originals so it only touches the text **above** the `\n---\n_-= Per source details. Do not edit below this line.=-_` token: the codelake attribution / first-catch prose is stripped from that top part, and everything from the token onward (the `## Source: amazon-inspector …` per-source details on the 11 records that have them) is preserved byte-for-byte. The 5 codelake first-catch records have no per-source section, so they're just the boilerplate removal.

Verified on the diffs and CI is green (validate / no-deletes / check-schema all pass).

calebbrown (approved):

PR #1511, opened 2026-09-10
Add OSV report: toga-ai@1.0.787 — unauthorized telemetry injection via postinstall

OGAREE:
Adds the malicious npm package `toga-ai@1.0.787`. Not previously in OSV; reported to npm Security on 2026-09-10.

It poses as a Claude Code AI harness but, by default, modifies the user's global `~/.claude/settings.json` during `npm install` through a postinstall hook. It injects 20 managed telemetry environment variables that enable Claude telemetry and content logging for user prompts, assistant responses, and tool execution details. It configures logs, metrics, and traces for the package-configured external collector at `https://otel.togaiq.com` using a hardcoded Bearer credential (`scripts/install.js:291`; credential omitted).

The script removes existing values for its managed telemetry keys before writing its defaults. Reinstalling therefore changes previously user-disabled values such as `OTEL_LOG_USER_PROMPTS=0` back to `1` (`scripts/install.js:359`). `TOGA_SKIP_TELEMETRY=1` prevents a later write but does not remove settings from an earlier installation. The package defines no uninstall lifecycle or other removal mechanism to restore the previous global settings.

The trigger is `package.json` → `node scripts/install.js --post-install`. `applyTelemetryEnv()` defines the hardcoded telemetry block at `scripts/install.js:287-307` and writes the global settings file through the path beginning at line 326. Because the target is the user-level Claude settings file, the configuration can affect subsequent supported Claude Code sessions outside the project in which the npm package was installed. Single affected version: `1.0.787`.

Analysis was static (code and data flow) on the published npm tarball; package code was not executed. Actual conversation transmission was not attempted or confirmed. SHA-256: `91b175cc05deefe1c672ce973b8a46b072df0eb91559a3eae03c962d338f3d86`.

### DIFF
--- a/osv/malicious/npm/toga-ai/MAL-0000-toga-ai.json
+++ b/osv/malicious/npm/toga-ai/MAL-0000-toga-ai.json
@@ -1,0 +1,40 @@
+{
+  "schema_version": "1.7.4",
+  "modified": "2026-09-10T00:00:00Z",
+  "published": "2026-09-10T00:00:00Z",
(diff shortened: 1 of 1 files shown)


calebbrown:
This does not look malicious. This behaviour appears to exist prior to the listed version.

Telemetry is not strictly malicious, although it is often unexpected and undesired.

Please provide a stronger justification for why this package should be considered malicious.

OGAREE:
Thank you for the review. After further investigation, the package appears
  to be an internal TOGA Technology team harness, and `otel.togaiq.com`
  appears to be part of the company’s AI infrastructure.

  The global configuration changes and overriding of disabled telemetry
  values remain concerning, but I do not have sufficient evidence of
  malicious intent or confirmed data exfiltration. I am therefore
  withdrawing this report.

  Thank you for your time and feedback

PR #1521, opened 2026-09-14
Add OSV report: faiss-cpu-avx512 (1 package, PyPI)

johncassil:
`faiss-cpu-avx512`@`1.9.0`-`1.9.7` on PyPI impersonates Meta's `faiss-cpu` and ships no FAISS code. Its `faiss_cpu_avx512/__init__.py` `_fetch_native()` downloads an xz-compressed binary from `https://files.catbox.moe/` (`ffoa50.xz` in 1.9.0-1.9.3, `0oreqo.xz` in 1.9.4, `vwfr94.xz` in 1.9.5, `3l3x0v.xz` in 1.9.6, `weu5u2.xz` in 1.9.7), writes it to `~/.cache/faiss_cpu_avx512/bin/indexwrm` with mode 0755, and runs it via `subprocess.Popen(start_new_session=True)`, restarting it every five seconds. It triggers on a call to the documented `warmup()`/`start()` entry point, not on install or import. The advertised `ARTIFACT_SHA256` integrity check is the literal string `PENDING_BUILD`, so the hash is never verified and the host's response is executed as-is. We are not claiming analysis of the fetched binary: we did not retrieve it. The wrapper's comments describe the binary as carrying a baked-in mining pool, wallet and relay and rewriting its process title to a `vllm serve` command line, and 1.9.4 onward start one worker per `nvidia-smi` GPU.

| Package | Versions | Host | First published |
|---|---|---|---|
| `faiss-cpu-avx512` | 1.9.0-1.9.7 | `files.catbox.moe` | 2026-09-09 |

Evidence: all eight wheels fetched from PyPI and read between 2026-09-11 and 2026-09-14; nothing executed, the binary host was not contacted. Finder: pkgwarden.

Related: none. No prior OSV/MAL record for this package as of 2026-09-14.


johncassil:
Superseding this with a combined report that adds four npm packages found in the same window (open-PR limit). New PR incoming with faiss-cpu-avx512 plus the npm entries.

PR #1527, opened 2026-09-16
Report malicious npm package jexkcode

Ridhdn:
Adds malicious versions of the npm package jexkcode (1.0.1 through 1.1.4).

The package is a WhatsApp library fork that, whenever a WhatsApp connection opens, automatically follows a hard-coded WhatsApp newsletter after a three-second delay. The call to newsletterFollow() is unconditional: it does not request user consent, expose a configuration option, or disclose this automatic account modification in the README.

Versions 1.0.1 through 1.1.1 contain the same automatic-follow logic but use a malformed newsletter JID, so the request may fail. Version 1.1.2 corrects the JID, enabling the subscription to succeed. Later commits remove the related failure and success logs; therefore, version 1.1.4 performs the account modification without visible output.

The trigger is the WhatsApp socket connection opening, after which the package invokes the auto-follow handler. The behavior is unrelated to the package’s stated functionality and modifies the user’s WhatsApp account without authorization.

Analysis was static on the published npm package versions and the corresponding public source history; package code was not executed. This report is limited to the unauthorized newsletter subscription behavior. Credential theft, data exfiltration, or remote code execution was not identified.


calebbrown (approved):

osv/malicious/npm/jexkcode/MAL-0000-jexkcode.json
```diff
@@ -0,0 +1,60 @@
+{
+  "schema_version": "1.7.4",
+  "id": "MAL-0000-jexkcode",
```
calebbrown: Please unset "id". This is populated by Workflows when the report is merged.

calebbrown (changes requested):
Thanks for the report! Just a small fix.

Ridhdn:
Removed the id field as requested. Thanks for the review.

calebbrown (approved):

calebbrown (approved):

PR #1532, opened 2026-09-17
feat(report): support GitHub Actions ecosystem and dual-identifier validation

Dwdr:
## Summary
Adds support for the `"GitHub Actions"` ecosystem in `ossf/malicious-packages`. This enables the ingestion, validation, and storage of malicious GitHub Actions advisories under `osv/malicious/github-actions/{owner}/{repo}/MAL-YYYY-NNNN.json`.

## Motivation & The Dual-Identifier Model
Unlike package registries (npm, PyPI) where published tarballs are immutable, GitHub Actions are Git repositories where tags can be force-pushed or mutated. To protect database integrity, this PR introduces a dual-identifier validation model:
1. **Exact Release Tags Only**: Reports may specify immutable point-in-time release tags (e.g. `v4.0.1`, `1.0.0`, `release-2026.01`).
2. **Rejection of Moving Tags**: Validation strictly rejects floating major/minor tags (e.g. `v4`, `4`, `v4.1`, `4.1`) and branch names (`main`, `master`, `latest`) in both version arrays and range events, as they convey no reliable point-in-time security guarantee.
3. **Direct Commit IDs (40/64-character SHAs)**: To handle unmerged imposter commits or compromised commits where maintainers moved the tag back to safety before an advisory was filed, researchers can submit the malicious 40-character commit SHA directly in `versions` or range events.

## Changes
* **`internal/report/validate.go`**:
  * Register `EcosystemGitHubActions = osvconstants.EcosystemGitHubActions` in `supportedEcosystems`.
  * Add `validateGitHubActionName()`: enforces canonical `{owner}/{repo}` (and composite subpath) naming, rejecting path traversals (`.`, `..`), empty segments, whitespace, and URL schemes.
  * Add `movingTagRegexes`, `isGitCommitSHA()`, and `isMovingTag()` heuristics to reject floating tags while accepting exact release tags and 40/64-char commit SHAs.
  * Integrate version and range event checks in `validateVulnInternal()` and `validateRange()`, supporting the OSV earliest-version sentinel introduced `"0"`.
* **`internal/report/report.go`**:
  * Update `canonicalizeName()` to strip trailing `.git` suffixes.
* **`internal/report/validate_test.go` & `internal/report/report_test.go`**:
  * Add 53 table-driven unit test subtests covering path generation, valid/invalid package names, exact tags, 40-char commit SHAs, and moving tag rejection.
  
## Verification
* Unit tests: `go test -v ./internal/report/...` (53 subtests PASS)
* Full repo test suite: `go test ./...` (100% PASS across all 15 packages)
* Linter & formatting: `gofmt -s -d` clean, `go vet ./...` clean (0 issues)


Dwdr:
/cc @calebbrown @elitsa-gosst 

Hi Caleb and Eli, opening this PR implementing Phase 2b (GitHub Actions ecosystem support and dual-identifier validation) as discussed. Whenever you have a chance, would appreciate your review!

internal/report/validate.go
```diff
@@ -178,9 +189,104 @@ func validatePackage(pkg *osvschema.Package) (osvconstants.Ecosystem, error) {
...
+		if err := validateGitHubActionName(name); err != nil {
+			return "", err
+		}
+	}
+
 	return ecosystem, nil
 }
 
+func validateGitHubActionName(name string) error {
+	if strings.TrimSpace(name) != name || strings.ContainsAny(name, " \t\r\n") {
```
elitsa-gosst: you're using this check several times across the file - consider making this a dedicated function instead

internal/report/validate.go
```diff
@@ -178,9 +189,104 @@ func validatePackage(pkg *osvschema.Package) (osvconstants.Ecosystem, error) {
...
+
+var movingTagRegexes = []*regexp.Regexp{
+	regexp.MustCompile(`^(?i)v?[0-9]+$`),
+	regexp.MustCompile(`^(?i)v?[0-9]+\.[0-9]+$`),
+	regexp.MustCompile(`^(?i)(latest|main|master|head|dev)$`),
+}
+
+// isGitCommitSHA reports whether s is a valid hex-encoded Git commit hash
+// (either 40-character SHA-1 or 64-character SHA-256).
+func isGitCommitSHA(s string) bool {
```
elitsa-gosst: we also have validateGitCommitID - maybe make the logic shared across the 2 functions?

internal/report/report.go
```diff
@@ -430,6 +430,9 @@ func canonicalizeName(name string, ecosystem osvconstants.Ecosystem) string {
 	case ecosystemGit:
 		// CanonForStorage is called for the name earlier during UnmarshalJSON
 		return name
+	case EcosystemGitHubActions:
+		name, _ = strings.CutSuffix(name, ".git")
```
elitsa-gosst, line 434: @calebbrown may have more comments about canonicalization

elitsa-gosst (changes requested):

internal/report/validate.go
```diff
@@ -35,12 +36,16 @@ const (
 	// ecosystemGit is synthetic ecosystem used when a git-based report is being
 	// processed, since those reports do not have any package data.
 	ecosystemGit = osvconstants.Ecosystem("Git")
+
+	// EcosystemGitHubActions is the ecosystem used for GitHub Actions reports.
+	EcosystemGitHubActions = osvconstants.EcosystemGitHubActions
```
calebbrown: My preference would be to not have a local copy of this. The ecosystemGit was necessary because it is a synthetic ecosystem that doesn't exist in osv-schema

internal/report/report.go
calebbrown: Yes. It should be lowercased - at least the `[org]/[repo]` part.

Are there many examples of the ".git" usage?

internal/report/report_test.go
```diff
@@ -34,8 +34,12 @@ import (
 )
 
 func testReport(ecosystem osvconstants.Ecosystem, name string) *report.Report {
-	rJSON := `{ "schema_version": "1.5.0", "summary": "test report", "affected": [{"package":{"ecosystem": "%s", "name": "%s"}, "versions": ["0"]}]}`
-	r, err := report.ReadJSON(bytes.NewBufferString(fmt.Sprintf(rJSON, ecosystem, name)))
+	ver := "0"
+	if ecosystem == report.EcosystemGitHubActions {
+		ver = "1.0.0"
+	}
```
calebbrown: What happens if "versions" is just "1.0.0" always?

(If it works nicely, then a comment explaining why it needs to be x.y.z for Github Actions would be good)

internal/report/validate.go
calebbrown: I also think validateGitCommitID is a bit easier to read and doesn't carry the magic "40" and "64" constants.

internal/report/validate.go
```diff
@@ -178,9 +189,104 @@ func validatePackage(pkg *osvschema.Package) (osvconstants.Ecosystem, error) {
...
+		}
+	}
+	return nil
+}
+
+var movingTagRegexes = []*regexp.Regexp{
+	regexp.MustCompile(`^(?i)v?[0-9]+$`),
+	regexp.MustCompile(`^(?i)v?[0-9]+\.[0-9]+$`),
+	regexp.MustCompile(`^(?i)(latest|main|master|head|dev)$`),
+}
```
calebbrown: Is it possible to do a search for how these are used, and if there are any counter-factuals?

e.g. repos with specific versions with only two digits, or v1-beta (as a floating tag) or any other variants?

internal/report/validate.go
```diff
@@ -178,9 +189,104 @@ func validatePackage(pkg *osvschema.Package) (osvconstants.Ecosystem, error) {
...
+
+// validateGitHubActionVersions ensures that versions listed for GitHub Actions
+// do not contain moving tags or invalid formatting.
+func validateGitHubActionVersions(versions []string) error {
+	for _, ver := range versions {
+		if ver == "" {
+			return fmt.Errorf("%w: version must not be empty", ErrInvalidOSV)
+		}
+		if strings.TrimSpace(ver) != ver || strings.ContainsAny(ver, " \t\r\n") {
```
calebbrown: " \t\r\n" is not exhaustive enough to cover all unicode whitespace. Also `strings.TrimSpace(ver) != ver` is unnecessary.

Consider replacing the entire line with:

`if strings.ContainsFunc(s, unicode.IsSpace) {`

internal/report/validate.go
calebbrown: This entire condition can be replaced with `strings.ContainsFunc(s, unicode.IsSpace)`.

It is also more correct.

internal/report/validate.go
```diff
@@ -178,9 +189,104 @@ func validatePackage(pkg *osvschema.Package) (osvconstants.Ecosystem, error) {
...
+	}
+
 	return ecosystem, nil
 }
 
+func validateGitHubActionName(name string) error {
+	if strings.TrimSpace(name) != name || strings.ContainsAny(name, " \t\r\n") {
+		return fmt.Errorf("%w: action name contains whitespace: %q", ErrInvalidOSV, name)
+	}
+	if strings.Contains(name, "://") || strings.HasPrefix(name, "git@") {
```
calebbrown: Might this validation be stronger if we assume a `[org]/[repo](/[extra...])` structure. GitHub has strong constraints on what characters can be in `[org]` and `[repo]` and if feels like that would be cleaner to validate.

In this case ":" and "@" would immediately be gone.

Also if we split on "/" then a "//" would result in an empty string being present `{ ... "" ... }` - which could trigger a failure.

Only a ".git" suffix might need to be explicitly stripped.

internal/report/validate.go
```diff
@@ -178,9 +189,104 @@ func validatePackage(pkg *osvschema.Package) (osvconstants.Ecosystem, error) {
...
+		if re.MatchString(trimmed) {
+			return true
+		}
+	}
+	return false
+}
+
+// validateGitHubActionVersions ensures that versions listed for GitHub Actions
+// do not contain moving tags or invalid formatting.
+func validateGitHubActionVersions(versions []string) error {
```
calebbrown, line 278: This function and validateGitHubActionRangeEvent essentially do the same thing.

Consider pulling out the functionality into a common validator for a single version.

internal/report/validate.go
Dwdr: Got it! I created a dedicated function with `strings.ContainsFunc(s, unicode.IsSpace)`

internal/report/validate.go
Dwdr: Got it, removed the function and directly check validateGitCommitID within isMovingTag()

internal/report/report.go
Dwdr: Got it! updated them to lowercased. 

For the ".git" usage, in real world reporting, .git creeps in when researchers copy clone URLs directly from the "Code" dropdown or when scanners extract remotes from .git/config and .gitmodules.

internal/report/validate.go
Dwdr: Ah I see, removed the local copy & switched to using `osvconstants.EcosystemGitHubActions`

internal/report/report_test.go
Dwdr: It works nicely! Setting `ver := "1.0.0"` unconditionally works for all existing ecosystems (npm, PyPI, Go, etc.) without breaking any tests across the repo.
I've removed the `if` check, set `ver := "1.0.0"` as the default, and added an explanatory comment (that GHA rejects 0 as a moving major tag)

internal/report/validate.go
Dwdr: I checked the 56 GHA advisories in the OSV dataset and GitHub's documentation:

Two-digit tags:
- 70 events use 3-digit SemVer, 10 use single/double-digit major tags (fixed: "41" in tj-actions, 6, 3), and only 1 used two digits (12.4 in wwbn/avideo, where it was their actual static release format).
- GitHub's official documentation specifically recommends moving major tags (v1), but does not officially document moving minor tags.

Prereleases & floating branch names: 
- Full SemVer prereleases represent immutable release milestones and are allowed 

To prevent false positives, I dropped the 2 digit tags so only single-component major tags are rejected, and added common floating branch names inside

internal/report/validate.go
Dwdr: Thanks! Validating against GitHub's rules will be much more robust than checking for ad-hoc strings

I've updated it to split on "/" (after stripping .git) and validate that segment 0 is a valid GitHub org/user and segment 1+ are valid repo/subpath characters. This'll eliminate colons, at-signs, and double slashes in one check.

internal/report/validate.go
Dwdr: Got it, I pulled it out into a common `validateGitHubActionVersion(val string, allowZero bool) error` helper that handles empty strings, whitespace, moving tags, and the optional "0" sentinel.

internal/report/validate.go
Dwdr: Got it! I replaced the checks with `strings.ContainsFunc(s, unicode.IsSpace) {` from the previous comment's dedicated function, hasWhiteSpace()

internal/report/validate.go
```diff
@@ -178,9 +192,104 @@ func validatePackage(pkg *osvschema.Package) (osvconstants.Ecosystem, error) {
...
+		return fmt.Errorf("%w: invalid github action org %q in %q", ErrInvalidOSV, org, name)
+	}
+	// Validate repo
+	repo := parts[1]
+	if !githubRepoRegex.MatchString(repo) || repo == "." || repo == ".." {
+		return fmt.Errorf("%w: invalid github action repo %q in %q", ErrInvalidOSV, repo, name)
+	}
+	// Validate optional subpath segments
+	for _, p := range parts[2:] {
+		if p == "" || p == "." || p == ".." || !githubRepoRegex.MatchString(p) {
```
calebbrown: Do sub-paths within a repo have the same constraints as repo names?

Given that org/repo/subaction is not used much I don't see this as an issue, but this may be more strict than needed (although not so bad I guess as it is easier to make it less strict than more strict)

internal/report/report.go
calebbrown: For the instances of `[org]/[repo]/[subaction]` is `"subaction"` case-insensitive as well?

internal/report/validate.go
```diff
@@ -178,9 +192,104 @@ func validatePackage(pkg *osvschema.Package) (osvconstants.Ecosystem, error) {
...
+		if err := validateGitHubActionVersion(ver, false); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// validateGitHubActionRangeEvent ensures that event values in GitHub Actions ranges
+// do not contain moving tags. The introduced event may be "0".
+func validateGitHubActionRangeEvent(val string, allowZero bool) error {
```
calebbrown: Consider removing this wrapping method and just call `validateGitHubActionVersion()` directly.

internal/report/validate.go
```diff
@@ -178,9 +192,104 @@ func validatePackage(pkg *osvschema.Package) (osvconstants.Ecosystem, error) {
...
+			return "", err
+		}
+	}
+
 	return ecosystem, nil
 }
 
+func hasWhitespace(s string) bool {
+	return strings.ContainsFunc(s, unicode.IsSpace)
+}
```
calebbrown: This wrapping method is not really needed, just call `strings.ContainsFunc(s, unicode.IsSpace)` directly.

The call is fairly obvious, and any ambiguity left is clarified by the error message following its use anyway.

internal/report/report.go
calebbrown: I think that rather than trimming the ".git" suffix here, in the validation method we should check that it does not exist at all in the repo component of the URL (i.e. parts[1]).

This is because using the ".git" suffix results in an error:

> Error: Unable to resolve action actions/checkout.git, repository not found

`canonicalizeName` assumes the name is already valid, and is intended to merely ensure that we don't end up with multiple reports for the same package under different variations of the same name.

calebbrown (changes requested):

internal/report/validate.go
Dwdr: Got it~ Removed the wrapper and called it directly. 

On the comment checking that ".git" suffix does not exist at all, I've moved the check into validateGitHubActionName(), with an ErrInvalidOSV error, and removed the trimming logic from `canonicalizeName`

internal/report/validate.go
Dwdr: Got it~ I removed it and called the func directly. 

Oh yes the subaction is case-sensitive (while [org]/[repo] is insensitive), since subaction maps to a physical filesystem path inside the cloned repository, which is case-sensitive on Linux runners.
I've updated canonicalizeName() to only lower case parts[0] and parts[1]

internal/report/validate.go
Dwdr: In composite actions (like slsa-framework/slsa-github-generator/actions/generator/generic), sub-paths represent directory names within the cloned repo, so validating each segment serves as a safe, conservative baseline (in which we can loosen it if unconventional path characters ever show up in real actions)
I removed p = "" as that's a dead code (the regex already enforces {1,100} min length).

